PR #7, opened 2024-08-26
move form tempdir to tempfile

Augustyniak:
[tempdir](https://github.com/rust-lang-deprecated/tempdir?tab=readme-ov-file) is deprecated, it's being developer as part of https://github.com/Stebalien/tempfile now.


murki (approved):
sure let's try. have we confirm that both use dirs in the same locations so it's fwd compatible?

PR #24, opened 2024-09-16
drop old logs during buffer upload

snowp:
Adds a runtime flag that allows specifying a threshold for how far back we go when we upload logs. These dropped logs will be gone forever and just serves as a simple way to allow bounding how far back we go when we flush the buffer.


bd-logger/src/consumer.rs
```diff
@@ -39,7 +41,10 @@ struct Flags {
   // The duration to wait before uploading an incomplete batch. The batch will be uploaded
   // at this point regardless of log activity.
   batch_deadline_watch: Watch<u32, bd_runtime::runtime::log_upload::BatchDeadlineFlag>,
-  // The interval at which to check the buffer for new records.
+
+  // The lookback window for the flush buffer uploads.
```
mattklein123, line 45: nit: can you say what unit this is? (It would be kind of cool if we had some type of helper which auto converted into a Duration, an idea for some other time.)

bd-logger/src/consumer.rs
```diff
@@ -601,9 +629,24 @@ impl CompleteBufferUpload {
     let mut total_logs = 0;
     loop {
       let entry = self.consumer.try_read();
+
       match entry {
         // Log available, add to batch and flush if batch size hit.
         Ok(log) => {
+          if let Ok(log) = root_as_log(&log) {
+            if let Some(lookback_window) = self.lookback_window {
+              if let Some(ts) = log.timestamp() {
```
mattklein123, line 638: Is it possible for there to ever not have a timestamp or is this an artifact of the fb API? Might be worth a comment.

mattklein123 (approved):
Nice LGTM

mattklein123 (approved):

PR #27, opened 2024-09-19
client stats: handle empty files as errors

mattklein123:
This can likely happen due to various failure conditions and this will correctly handle that case and perform relevant cleanup.

Fixes BIT-3792


bd-client-stats/src/stats.rs
```diff
@@ -87,6 +87,15 @@ pub trait SerializedFileSystem: Sync {
     // For now we just read the entire thing and then decompress it in memory. We can consider
     // streaming later.
     let compressed_bytes = self.read_file(path).await?;
+
+    // We should never write empty files. If there is no data this was a partial write, full disk
+    // issue, or some other problem.
+    // TODO(mattklein123): Compression gives some protection against further corruption but we
```
snowp, line 93: It would probably be worth doing some generic protobuf file writing logic that always handles the crc header, as I'm sure our config persistence is also susceptible to file corruption

bd-client-stats/src/stats.rs
```diff
@@ -87,6 +87,15 @@ pub trait SerializedFileSystem: Sync {
...
+    // should also be writing and reading some type of CRC has part of this process.
+    if compressed_bytes.is_empty() {
+      return Err(anyhow::anyhow!("unexpected empty file"));
```
snowp, line 96: nit: anyhow::bail!

bd-client-stats/src/stats.rs
```diff
@@ -191,25 +200,33 @@ impl<F: SerializedFileSystem> Uploader<F> {
...
+          );
+          Some(pending_request)
+        },
+        Err(e) => {
+          // We failed to read the data, so the file must be bad. This could happen if we change
+          // the schema in an incompatible way or if the file is corrupt. Delete the file and
+          // accept the loss of this upload.
+
+          // TODO(snowp): Technically we could end up in a situation here where we are stuck
+          // trying to delete a bad file which then spams the error handler.
```
snowp, line 221: We can remove this at this point I think since we limit how many times we report an error deeper in the error handler

bd-client-stats/src/stats.rs
```diff
@@ -226,14 +243,19 @@ impl<F: SerializedFileSystem> Uploader<F> {
...
-      log::warn!("failed to read aggregated stats file due to data corruption, deleting file");
-      let _ignored = self.fs.delete_file(&*AGGREGATED_STATS_FILE).await;
-      return Ok(());
+    {
+      Ok(stats) => stats,
+      Err(e) => {
+        log::warn!(
+          "failed to read aggregated stats file due to data corruption, deleting file: {e}"
+        );
+        let _ignored = self.fs.delete_file(&*AGGREGATED_STATS_FILE).await;
```
snowp, line 256: Don't remember why we ignore the error here but not in the other delete_file call but seems a bit strange

snowp (approved):

bd-client-stats/src/stats.rs
mattklein123: Yes agreed. I can look at this as a follow up.

bd-client-stats/src/stats.rs
mattklein123: I'm not sure reporting failure to delete is useful so maybe I will just remove the other one.

mattklein123: In looking at this the error handling here is all over the place. Let me try to clean it up.

mattklein123:
@snowp updated per comments. Let me know what you think.

snowp (approved):

PR #34, opened 2024-09-27
api: add the ability to remotely shutoff clients

mattklein123:
This is known as the "client kill" mechanism. There are two different paths, both leading to the client being disabled with no network connection to the SaaS for some period of time:
1. The client receives an unauthenticated response.
2. The client receives a specific runtime generic kill update which can be more targeted.

The client will persist a file with a wakeup time after which the client will come back online and attempt to contact the server again.

Fixes BIT-3577

### DIFF
--- a/bd-api/src/api.rs
+++ b/bd-api/src/api.rs
@@ -27,12 +27,13 @@ use anyhow::anyhow;
 use backoff::backoff::Backoff;
 use backoff::exponential::{ExponentialBackoff, ExponentialBackoffBuilder};
 use backoff::SystemClock;
+use bd_client_common::file::{read_compressed_protobuf, write_compressed_protobuf};
+use bd_client_common::zlib::DEFAULT_MOBILE_ZLIB_COMPRESSION_LEVEL;
 use bd_client_stats_store::{Counter, CounterWrapper, Scope};
 use bd_grpc_codec::{
   Compression,
   Encoder,
   OptimizeFor,
-  DEFAULT_MOBILE_ZLIB_COMPRESSION_LEVEL,
   GRPC_ACCEPT_ENCODING_HEADER,
   GRPC_ENCODING_DEFLATE,
   GRPC_ENCODING_HEADER,
@@ -47,6 +48,7 @@ use bd_proto::protos::client::api::{
   handshake_response,
   ApiRequest,
   ApiResponse,
+  ClientKillFile,
   ConfigurationUpdateAck,
   HandshakeRequest,
   PingRequest,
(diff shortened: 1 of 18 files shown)


mattklein123:
I just realized that during initial development this is going to be annoying. Maybe we need to make the warning more explicit? Let me know what you think.

bd-api/src/api.rs
```diff
@@ -377,9 +406,88 @@ impl Api {
...
+    Ok(Some(OffsetDateTime::from_unix_timestamp(kill_until)?))
+  }
+
+  async fn write_kill_file(&mut self, duration: Duration) -> anyhow::Result<()> {
+    self.client_killed = true;
+
+    // In order to have basic protection around IO errors we will zlib "compress" the kill file
+    // primarily as a free way to get the CRC.
+    let kill_until = self.time_provider.now() + duration;
+    let mut encoder = ZlibEncoder::new(Vec::new(), flate2::Compression::new(3));
```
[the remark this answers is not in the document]
mattklein123: Sure will do.

bd-api/src/api.rs
```diff
@@ -377,9 +405,103 @@ impl Api {
...
+          "kill file exists and is still active ({} remaining), killing client",
+          kill_until_time - self.time_provider.now()
+        );
+        log::warn!(
+          "Attention: The SDK has been force disabled due to either a previous authentication \
+           failure or a remote server configuration. Double check your API key or contact support."
+        );
+        self.client_killed = true;
+      } else {
+        // Delete the kill file if the kill duration has passed. This will allow the client to
```
snowp, line 458: Maybe note that we also wipe the file if the API key changes?

bd-client-common/src/file.rs
```diff
@@ -0,0 +1,41 @@
...
+  // issue, or some other problem. We use zlib for compression which includes a CRC at the end
+  // so as long as the file is not empty we can be sure that the data is not corrupted.
+  if compressed_bytes.is_empty() {
+    anyhow::bail!("unexpected empty file");
+  }
+
+  // The files are likely not large enough to deal with streaming decompression on top of flate2.
+  // For now we just read the entire thing and then decompress it in memory. We can consider
+  // streaming later.
+  let mut decoder = ZlibDecoder::new(compressed_bytes);
```
snowp, line 37: This is probably old code but can we pass a Reader to the decoder so that we don't have to pull the whole file contents into memory? 

Same above /w using write_to_writer instead of `write_to_bytes`

snowp (approved):
Couple small comments but overall LGTM

bd-client-common/src/file.rs
mattklein123: The problem ends up being async vs not. We would need an adapter to use this code from async and I don't think it's worth it given the comment here about file sizes not being very large. I think we could fix this in the future if we ever deal with larger files.

bd-api/src/api.rs
mattklein123: Will do.

PR #45, opened 2024-10-17
client stats: rewrite aggregation to support multiple windows

mattklein123:
This is a substantial rewrite of the client stats local aggregation functionality. The code now maintains up to N files, each of a maximum aggregation duration of M minutes. This makes sure that during periods of offline connecticity we maintain discrete metric windows. Further, when devices are offline for a long time and then come online again, new metrics will not get mixed with old metrics.

Fixes BIT-3953
Fixes BIT-2850


bd-client-stats/src/stats.rs
```diff
@@ -9,347 +9,127 @@
...
-  ) -> Self {
-    Self {
-      shutdown,
-      upload_interval_flag,
-      data_flush_tx,
-      fs,
+impl Ticker for RuntimeWatchTicker {
+  async fn tick(&mut self) {
+    if self.interval.is_none() {
+      self.interval = Some(self.receiver.borrow_and_update().jittered_interval_at());
```
snowp, line 65: So does this mean that we are jittering all the intervals? Which would mean that if we set the interval to say 2m the average interval is more like 1m (depending on the jitter distribution)? If anything might be nice to have this as part of the doc string on the runtime flag

bd-client-stats/src/stats.rs
```diff
@@ -485,47 +276,125 @@ impl<T: TimeProvider, F: SerializedFileSystem> Flusher<T, F> {
...
+          .file_manager
+          .complete_pending_upload(&upload_response.uuid)
+          .await,
+        "complete pending upload",
+      );
+      // During startup or after getting network connectivity back it's possible that we will have
+      // a number of pending uploads to process. Go ahead and see if we have an old file at the
+      // head of the list which we should upload immediately.
+      // TODO(mattklein123): It would be better to batch all of the "old" files into a single
+      // upload request. We can do this in the future.
```
snowp, line 362: If we are allocating a UUID per upload and use this for idempotency this might get tricky? Or would these pending uploads never have been attempted uploaded?

bd-client-stats/src/file_manager.rs
```diff
@@ -0,0 +1,424 @@
...
+      max_aggregation_window_per_file: runtime_loader.register_watch()?,
+    })
+  }
+
+  // Read an existing snapshot from disk to merge into, or create a new one.
+  pub async fn get_or_create_snapshot(&self) -> anyhow::Result<StatsUploadRequestHandle> {
+    let mut inner = self.inner.lock().await;
+    let initialized_inner = inner.get_initialized().await?;
+
+    let use_existing = if initialized_inner.index.back().map_or(true, |file| {
```
snowp, line 233: This block is a bit hard to read, maybe pull the "create new snapshot check" out into a var instead of this big if condition?

bd-client-stats/src/file_manager.rs
```diff
@@ -0,0 +1,424 @@
...
+      {
+        log::debug!("snapshot is too old, creating new snapshot");
+        true
+      } else {
+        false
+      }
+    }) {
+      if self.max_aggregated_files.read() <= u32::try_from(initialized_inner.index.len()).unwrap() {
+        log::debug!("max files reached, popping oldest snapshot");
+        initialized_inner.delete_oldest_snapshot().await?;
```
snowp, line 249: obviously a bit hard to emit stats within the stats system but would be nice to have a way to track how often we end up expiring snapshots

bd-client-stats/src/file_manager.rs
```diff
@@ -0,0 +1,424 @@
...
+    &self,
+    mut handle: StatsUploadRequestHandle,
+    mut snapshot: Snapshot,
+  ) -> anyhow::Result<()> {
+    let mut inner = self.inner.lock().await;
+    let initialized_inner = inner.get_initialized().await?;
+
+    // Even for a pre-existing snapshot, we are being passed back a new snapshot so we need to
+    // initialized the period start again.
+    snapshot.occurred_at = Some(Occurred_at::Aggregated(Aggregated {
```
snowp, line 324: I wonder if we should be storing both start/end of the snapshot and send both? Especially with the new jitter I think we'll end up with even more variable snapshot periods so maybe the server should insert them at avg(start, end) instead of always using the start?

bd-client-stats/src/stats.rs
```diff
@@ -485,47 +276,125 @@ impl<T: TimeProvider, F: SerializedFileSystem> Flusher<T, F> {
...
+        .file_manager
+        .get_or_create_pending_upload(only_if_file_is_old)
+        .await?
+      {
+        return flusher.process_pending_upload(pending_upload).await;
+      }
+      Ok(None)
+    }
+
+    // Note on error handling: while we could probably gracefully handle some of the failing I/O
```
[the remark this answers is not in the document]
mattklein123: Yeah I've seen that. I'm not really sure what to do better here though. What I don't know is if this new implementation will go haywire in other ways like leaking files, etc.

bd-client-stats/src/stats.rs
mattklein123: No this just jitters the start of the interval. I will add a comment.

bd-client-stats/src/stats.rs
mattklein123: The idea here would be to collapse the old files into a single upload and track it, but we have idempotency issues throughout the stack unfortunately, see all the TODOs in the SaaS. This is a really messy issue.

bd-client-stats/src/file_manager.rs
mattklein123: Will do.

bd-client-stats/src/file_manager.rs
mattklein123: Yeah not sure how to do this. I guess I can add an internal stat but it might not be very accurate.

bd-client-stats/src/file_manager.rs
mattklein123: See above re: jitter, but having end makes sense to me. I can add that to the proto.

mattklein123:
Updated per first round of comments. We can discuss more when @Augustyniak is back next week.

PR #47, opened 2024-10-19
grpc: reliability functionality

mattklein123:
1) Allow HTTP/2 stream settings to be configured.
2) Add generic read stop functionality to streaming APIs. This allows
   callers to use a watch to stop reading from streams and apply back
   pressure.

### DIFF
--- a/bd-grpc/src/lib.rs
+++ b/bd-grpc/src/lib.rs
@@ -52,7 +52,7 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 use time::ext::NumericalDuration;
 use time::Duration;
-use tokio::sync::{mpsc, Semaphore};
+use tokio::sync::{mpsc, watch, Semaphore};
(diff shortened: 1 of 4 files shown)


mattklein123:
Needs clippy fixes from other PRs so putting in draft for now.

bd-grpc/src/lib.rs
```diff
@@ -501,9 +501,11 @@ impl<C: Connect + Clone + Send + Sync + 'static> Client<C> {
       },
     }
 
+    log::debug!("sending streaming request");
```
snowp, line 504: did you want to leave these logs in? Not sure how useful they would be in general

snowp (approved):

bd-grpc/src/lib.rs
mattklein123: It would just for me debugging. I can remove them.

PR #52, opened 2024-10-24
add session replay target, wire up take screenshot action

Augustyniak:
Fixes BIT-3972

1. Wire up screenshot take action to `session_replay::Target` that's exposed on the public interface of the Rust logger.
2. Move the control over whether periodic captured of session replay wireframes are performed or not to the Rust layer itself. The public interface to control the interval of session replay interval will be removed, instead we will control is with remove runtime updates.

### DIFF
--- a/bd-logger/src/log_replay.rs
+++ b/bd-logger/src/log_replay.rs
@@ -17,7 +17,7 @@ use bd_matcher::buffer_selector::BufferSelector;
 use bd_runtime::runtime::filters::FilterChainEnabledFlag;
 use bd_runtime::runtime::workflows::WorkflowsEnabledFlag;
 use bd_runtime::runtime::{BoolWatch, ConfigLoader};
-use bd_session_replay::TakeScreenshotHandler;
+use bd_session_replay::CaptureScreenshotHandler;
 use bd_workflows::actions_flush_buffers::BuffersToFlush;
 use bd_workflows::engine::{WorkflowsEngine, WorkflowsEngineConfig};
 use std::borrow::Cow;
@@ -89,7 +89,7 @@ pub struct ProcessingPipeline {
 
   trigger_upload_tx: Sender<TriggerUpload>,
   buffers_to_flush_rx: Option<Receiver<BuffersToFlush>>,
-  take_screenshot_handler: TakeScreenshotHandler,
+  capture_screenshot_handler: CaptureScreenshotHandler,
 
   workflows_enabled_flag: BoolWatch<WorkflowsEnabledFlag>,
   filter_chain_enabled_flag: BoolWatch<FilterChainEnabledFlag>,
@@ -106,7 +106,7 @@ impl ProcessingPipeline {
     flush_buffers_tx: Sender<BuffersWithAck>,
     flush_stats_trigger: Option<FlushTrigger>,
     trigger_upload_tx: Sender<TriggerUpload>,
-    take_screenshot_handler: TakeScreenshotHandler,
+    capture_screenshot_handler: CaptureScreenshotHandler,
 
     config: ConfigUpdate,
 
@@ -153,7 +153,7 @@ impl ProcessingPipeline {
       trigger_upload_tx,
       buffers_to_flush_rx,
 
-      take_screenshot_handler,
+      capture_screenshot_handler,
 
       workflows_enabled_flag,
       filter_chain_enabled_flag,
@@ -256,8 +256,8 @@ impl ProcessingPipeline {
         );
       }
 
-      if result.take_screenshot {
-        self.take_screenshot_handler.take_screenshot();
+      if result.capture_screenshot {
+        self.capture_screenshot_handler.capture_screenshot();
       }
 
       Self::write_to_buffers(
--- a/bd-runtime/src/runtime.rs
+++ b/bd-runtime/src/runtime.rs
@@ -751,14 +751,14 @@ pub mod session_replay {
   use time::ext::NumericalDuration as _;
 
   bool_feature_flag!(
-    PeriodicWireframesEnabledFlag,
-    "session_replay.periodic_wireframes.enabled",
+    PeriodicScreensEnabledFlag,
+    "session_replay.screens.enabled",
     false
   );
 
   duration_feature_flag!(
     ReportingIntervalFlag,
-    "session_replay.periodic_wireframes.interval_ms",
+    "session_replay.screens.interval_ms",
     3.seconds()
   );
 
--- a/bd-session-replay/src/lib.rs
+++ b/bd-session-replay/src/lib.rs
@@ -34,21 +34,21 @@ fn test_global_init() {
 // An interface implementing the act of capturing user screens.
 pub trait Target {
   // Instruct the target to capture a privacy-preserving and bandwidth-efficient representation
-  // of the user's screen. The target should capture the wireframe and send it using the
-  // `logger::log_session_replay_wireframe` method. The target is expected to operate
+  // of the user's screen. The target should capture the screen and send it using the
+  // `logger::log_session_replay_screen` method. The target is expected to operate
   // asynchronously, as accessing the application view hierarchy requires executing on the main
-  // thread, and the `capture_wireframe` method is always called from a non-main thread.
-  fn capture_wireframe(&self);
+  // thread, and the `capture_screen` method is always called from a non-main thread.
+  fn capture_screen(&self);
 
   // Instruct the target to capture a pixel-perfect representation of the user's screen.
-  // The target should capture the wireframe and send it using the
-  // `logger::log_session_screenshot` method. The target is expected to operate asynchronously,
-  // as accessing the application view hierarchy requires executing on the main thread, and the
-  // `take_screenshot` method is always called from a non-main thread.
+  // The target should capture the screenshot and send it using the
+  // `logger::log_session_replay_screenshot` method. The target is expected to operate
+  // asynchronously, as accessing the application view hierarchy requires executing on the main
+  // thread, and the `take_screenshot` method is always called from a non-main thread.
   //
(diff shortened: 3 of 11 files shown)


bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+          session_replay::PeriodicWireframesEnabledFlag::register(runtime_loader).unwrap(),
+        reporting_interval_flag: session_replay::ReportingIntervalFlag::register(runtime_loader)
+          .unwrap(),
+        is_screenshotting_enabled_flag,
+      },
+      take_screenshot_tx,
+    )
+  }
+
+  fn create_interval(interval: Duration, fire_immediately: bool) -> tokio::time::Interval {
```
mattklein123, line 88: Prefer using time::Duration, then below you can use the interval() and interval_at() helpers in bd_time

bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+      self.reporting_interval = Some(Self::create_interval(self.reporting_interval_rate, true));
+    }
+
+    let local_shutdown = shutdown.cancelled();
+    tokio::pin!(local_shutdown);
+
+    loop {
+      tokio::select! {
+        () = async {
+          if let Some(reporting_interval) = &mut self.reporting_interval {
```
mattklein123, line 119: Just unwrap in here, the outer guard makes it impossible to get in here without it being is_some()

bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+            log::debug!("session replay recorder taking screenshot");
+            self.target.take_screenshot();
+          }
+        },
+        _ = self.is_periodic_reporting_enabled_flag.changed() => {
+          self.is_periodic_reporting_enabled = self.is_periodic_reporting_enabled_flag.read();
+        },
+        _ = self.is_screenshotting_enabled_flag.changed() => {
+          self.is_screenshotting_enabled = self.is_screenshotting_enabled_flag.read();
+        },
```
mattklein123, line 145: Pretty sure this is going to spin. You want read_mark_update(). Same elsewhere.

bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+
+//
+// Target
+//
+
+pub trait Target {
+  fn capture_wireframes(&self);
+
+  fn take_screenshot(&self);
+}
```
mattklein123, line 33: How is the platform code going to work here? Is this going to block or be async in some way? Can you add some comments? If it's async we come back to my other question about spamming the channel.

mattklein123 (changes requested):
Nice, some comments, thanks.

bd-runtime/src/runtime.rs
```diff
@@ -747,6 +747,28 @@ pub mod resource_utilization {
...
+  bool_feature_flag!(
+    ScreenshotsEnabledFlag,
+    "session_replay.screenshots.enabled",
+    false
+  );
+
+  duration_feature_flag!(
+    ReportingIntervalFlag,
+    "session_replay.periodic_wireframes.reporting_interval_ms",
+    3.seconds()
```
murki, line 768: ms or seconds?

murki: also currently this is passed by the user during start(), are you proposing to remove that capability altogether and only be able to control it via runtime var?

bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+
+impl Recorder {
+  pub fn new(
+    target: Box<dyn Target + Send + Sync>,
+    runtime_loader: &Arc<ConfigLoader>,
+  ) -> (Self, Sender<()>) {
+    // Limit the buffer size to 1 to reduce the risk of putting too much pressure on the
+    // application's main thread when dequeuing the screenshot actions from the channel and
+    // passing them to the platform layer, which performs the actual screenshotting on the main
+    // thread.
```
mattklein123, line 73: Sorry for possibly asking the obvious, but will the platform layer block until the screen shot is taken? If it spawns some other async task this would still be able to loop unbounded.

bd-session-replay/src/lib.rs
mattklein123: The two functions below are exactly `interval()` and `interval_at()` unless I am missing something: https://github.com/bitdriftlabs/shared-core/blob/main/bd-time/src/lib.rs#L130-L136

bd-session-replay/src/lib.rs
mattklein123: ping on this

bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+            reporting_interval.tick().await;
+          }
+        }, if self.reporting_interval.is_some() && self.is_periodic_reporting_enabled => {
+          log::debug!("session replay recorder capturing wireframe");
+          self.target.capture_wireframe();
+        },
+        _ = self.reporting_interval_flag.changed() => {
+          self.reporting_interval = Some(
+            Self::create_interval(
+              self.reporting_interval_flag.read().unsigned_abs(),
```
mattklein123, line 147: read_mark_update

bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    loop {
+      tokio::select! {
+        () = async {
+          if let Some(reporting_interval) = &mut self.reporting_interval {
+            reporting_interval.tick().await;
+          }
+        }, if self.reporting_interval.is_some() && self.is_periodic_reporting_enabled => {
+          log::debug!("session replay recorder capturing wireframe");
+          self.target.capture_wireframe();
```
mattklein123, line 142: This one is less of a risk but technically if we get backed up we could be issuing N of these. Per my other question not sure how the platform layer is going to work but worth a comment around this if it isn't an issue.

bd-session-replay/src/lib.rs
```diff
@@ -133,35 +156,49 @@ impl Recorder {
...
-            reporting_interval.tick().await;
-          }
-        }, if self.reporting_interval.is_some() && self.is_periodic_reporting_enabled => {
+        () = async { self.reporting_interval.as_mut().unwrap().tick().await; },
+          if self.reporting_interval.is_some() && self.is_periodic_reporting_enabled => {
           log::debug!("session replay recorder capturing wireframe");
           self.target.capture_wireframe();
         },
-        _ = self.reporting_interval_flag.changed() => {
+        _ = self.reporting_interval_rate_flag.changed() => {
```
mattklein123, line 168: As an aside, as a follow up we might consider removing `read_mark_update` and having `changed` return the new value. This would be a lot less fragile and all the typed implement Copy so there is no perf issue.

bd-session-replay/src/lib.rs
```diff
@@ -170,3 +207,30 @@ impl Recorder {
...
+    _fields: &mut bd_log_primitives::AnnotatedLogFields,
+  ) {
+    if !(log_type == LogType::Replay && msg.as_str() == Some(SESSION_REPLAY_SCREENSHOT_LOG_MESSAGE))
+    {
+      return;
+    }
+
+    if self.next_screenshot_tx.try_send(()).is_err() {
+      log::debug!("failed to send ready for next screenshot signal");
+    }
```
mattklein123, line 238: This should probably be an unexpected error?

bd-logger/src/log_replay.rs
```diff
@@ -251,6 +254,12 @@ impl ProcessingPipeline {
         );
       }
 
+      if result.take_screenshot {
+        if let Err(e) = self.take_screenshot_tx.try_send(()) {
```
mattklein123, line 258: Now that you have the concurrency control it's probably fine to use normal send here and not try send? In practice it probably doesn't matter but if you keep try_send I would add a comment on why.

bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+  bd_test_helpers::test_global_init();
+}
+
+//
+// Target
+//
+
+// An interface implementing the act of capturing user screens.
+pub trait Target {
+  // Instrument the target to capture a privacy-preserving and bandwidth-efficient representation
```
mattklein123, line 34:
```suggestion
  // Instruct the target to capture a privacy-preserving and bandwidth-efficient representation
```

? Same below.

mattklein123 (approved):
LGTM with small comment, nice!

bd-logger/src/log_replay.rs
```diff
@@ -254,11 +256,7 @@ impl ProcessingPipeline {
         );
       }
 
-      if result.take_screenshot {
-        if let Err(e) = self.take_screenshot_tx.try_send(()) {
-          log::debug!("failed to send take screenshot signal: {e}");
-        }
-      }
+      self.take_screenshot_handler.take_screenshot();
```
mattklein123, line 259: Don't you still need to guard this with result.take_screenshot? Seems like there should be some test to catch this?

bd-session-replay/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+  take_screenshot_tx: Sender<()>,
+}
+
+impl TakeScreenshotHandler {
+  pub fn take_screenshot(&self) {
+    // We use a channel with a capacity of one and employ `try_send` to avoid waiting for space in
+    // the channel before sending a new signal. This is especially important as the method is
+    // expected to be called on logging hot path.
+    // The channel may become full if "take screenshot" requests arrive faster than the platform
+    // layer can process them. In such cases, new requests are ignored.
```
mattklein123, line 230: Ah ok yeah this makes sense. Forget this is going from the logging code to the recorder.

PR #57, opened 2024-11-06
filters: allow extractions from log message

Augustyniak:


bd-log-filter/src/filter_chain_test.rs
```diff
@@ -433,6 +433,29 @@ fn invalid_regex_match() {
   assert!(filter_chain.filters.is_empty());
 }
 
+#[test]
+fn extracts_message_portion_and_creates_field() {
+  let filter_chain = make_filter_chain(
```
snowp, line 438: wanna do similar coverage for log_field and log_level?

snowp (approved):

PR #60, opened 2024-11-12
sankey: fix intent negotiation process

Augustyniak:
The clue of the changes is https://github.com/bitdriftlabs/shared-core/pull/60/files#diff-30cf2cdd0f9170897ba93256c5490824715193fcd9a7da75884932ff95d7f213R849 but I also modified api protos in https://github.com/bitdriftlabs/api/pull/34 and did a follow up refactor of how we handle `IntentResponse`/`IntentDecision` in code.

### DIFF
--- a/bd-api/src/api.rs
+++ b/bd-api/src/api.rs
@@ -848,7 +848,7 @@ impl Api {
         },
         Some(ResponseKind::SankeyPathUploadIntent(sankey_path_upload_intent)) => {
           log::debug!(
-            "received ack for sankey path intent upload {:?}, decision: {:?}",
+            "received ack for sankey path upload intent {:?}, decision: {:?}",
             sankey_path_upload_intent.intent_uuid,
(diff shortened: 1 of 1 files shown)


bd-workflows/src/sankey_diagram.rs
```diff
@@ -197,24 +205,27 @@ impl Processor {
...
 
     if let Err(error) = response.await {
       log::debug!("failed to wait for sankey upload response: {error}");
-      self.stats.upload_completion_failure_total.inc();
+      self.stats.upload_completion_failures.inc();
       return;
     }
 
-    self.stats.upload_completion_success_total.inc();
+    // Augustyniak: Is it risky to insert the path only in the case of a successful upload? maybe we
```
snowp, line 218: What's the consequence of this? Are you worried that we would keep sending intents or is that covered elsewhere?

snowp (approved):

PR #96, opened 2025-02-11
add device_id as a matching field

snowp:
Adds _device_id as a matching field on all logs. This should make it possible to write a matcher against device IDs without us having to upload the ID for each log

BIT-4565

### DIFF
--- a/bd-logger/src/device_id.rs
+++ b/bd-logger/src/device_id.rs
@@ -1,3 +1,10 @@
+// shared-core - bitdrift's common client/server libraries
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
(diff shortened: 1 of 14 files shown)


bd-logger/src/device_id.rs
```diff
@@ -0,0 +1,43 @@
...
+    _msg: &LogMessage,
+    _fields: &mut AnnotatedLogFields,
+    matching_fields: &mut AnnotatedLogFields,
+  ) {
+    // The device_id field is populated by lapi on upload based on the handshake, so by putting
+    // it in the matching fields it will be matchable and also present on all logs without us
+    // having to add an explicit field that eats up buffer capacity.
+    matching_fields.push(AnnotatedLogField::new_ootb(
+      "_device_id".to_string(),
+      self.device_id.clone().into(),
```
mattklein123, line 40: For stuff like this that never changes, I wonder if we could support Arc or Cow to avoid all these copies for no reason? Might not be that hard to support different types in an enum...

mattklein123 (approved):

PR #105, opened 2025-02-26
add an initial crash handling implementation

snowp:
This adds a basic crash handling implementation that expects the platform layer to present raw crash reports in a specific directory. These reports are consumed by the handler and emitted as crash logs into the logger

To help the platform layer know where to look for crash reports, a runtime flag is used to populate a file on disk that the platform layer can use to determine which directories to look for crashes


bd-crash-handler/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    Self {
+      crash_directories_flag,
+      crash_directory: sdk_directory.join(CRASH_DIRECTORY),
+      crash_logger,
+      shutdown,
+    }
+  }
+
+  pub async fn try_ensure_directories_exist(&self) -> anyhow::Result<()> {
```
murki, line 77: so who calls this?

snowp: should be called during startup by rust

snowp:
This PR is pretty big since i refactored some of the integration tests to be easier to split up, ignore changes to logger_integration and embedded, they are just moved around. The setup.rs is split out from logger_integration

snowp:
I have read the CLA Document and I hereby sign the CLA

FranAguilera (approved):
I think changes for string runtime makes sense

PR #106, opened 2025-02-28
improve ergonomics when matching against uploaded logs in test

snowp:
Instead of using a number of free functions that take &[u8], instead we have the log upload test helpers return a newtype that wraps the data. This allows us to move all the helpers to take &self which makes them easier to discover via IDEs and the code a bit easier to follow

### DIFF
--- a/bd-buffer/src/ring_buffer_test.rs
+++ b/bd-buffer/src/ring_buffer_test.rs
@@ -169,7 +169,7 @@ async fn test_ring_buffer_manager() {
 }
 
 #[tokio::test]
-async fn test_ring_buffer_stats() {
+async fn ring_buffer_stats() {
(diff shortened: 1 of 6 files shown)


bd-buffer/src/ring_buffer_test.rs
```diff
@@ -250,7 +250,7 @@ async fn test_ring_buffer_stats() {
   );
 
   // Write a bunch of entries that will overflow the small buffers
-  for _ in 0 .. 1000 {
+  for _ in 0 .. 5000 {
```
snowp, line 253: this test seemed a bit flaky so i just made it a bit more aggressive to see if that makes a difference

FranAguilera (approved):

PR #107, opened 2025-03-05
workflows: implement generate log action

mattklein123:
Also remove the enable workflow runtime key.

### DIFF
--- a/bd-logger/src/async_log_buffer.rs
+++ b/bd-logger/src/async_log_buffer.rs
@@ -18,11 +18,13 @@ use crate::metadata::MetadataCollector;
 use crate::network::{NetworkQualityInterceptor, SystemTimeProvider};
 use crate::pre_config_buffer::PreConfigBuffer;
 use crate::{internal_report, network};
+use anyhow::anyhow;
 use bd_buffer::BuffersWithAck;
 use bd_client_common::error::{handle_unexpected, handle_unexpected_error_with_details};
-use bd_log_metadata::{AnnotatedLogFields, MetadataProvider};
+use bd_log_metadata::{AnnotatedLogFields, LogFieldKind, MetadataProvider};
 use bd_log_primitives::{
   log_level,
+  AnnotatedLogField,
   Log,
   LogField,
   LogFieldValue,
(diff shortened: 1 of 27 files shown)


bd-logger/src/async_log_buffer.rs
```diff
@@ -490,23 +529,21 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
...
-    }
+      LoggingState::Initialized(initialized_logging_context) => self
+        .replayer
+        .replay_log(
+          log,
+          log_processing_completed_tx,
+          &mut initialized_logging_context.processing_pipeline,
+        )
+        .await
+        .map_err(|e| anyhow!("failed to replay async log buffer log: {e}"))?,
```
snowp, line 543: since you're here this could probably be using `context()` instead

bd-logger/src/async_log_buffer.rs
```diff
@@ -680,6 +717,7 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
     session_id: String,
     occurred_at: time::OffsetDateTime,
   ) -> anyhow::Result<()> {
+    // TODO(mattklein123): Should we support injected logs for internal logs?
```
snowp, line 720: I think this just puts logs on the async queue so it should just work?

bd-workflows/src/generate_log.rs
```diff
@@ -0,0 +1,137 @@
...
+
+fn resolve_reference<'a>(
+  extractions: &'a TraversalExtractions,
+  reference: &'a ValueReference,
+  current_log_fields: &'a FieldsRef<'a>,
+) -> Option<StringOrFloat<'a>> {
+  match reference.value_reference_type.as_ref()? {
+    Value_reference_type::Fixed(value) => Some(StringOrFloat::String(value)),
+    Value_reference_type::FieldFromCurrentLog(field_name) => {
+      // TODO(mattklein123): Should we also look in matching fields?
```
snowp, line 48: probably yes

bd-logger/src/async_log_buffer.rs
mattklein123: OK let me do that.

bd-logger/src/async_log_buffer.rs
mattklein123: Let me double check I think this is direct and not via the queue so it needs to actually handle the injected logs returned.

bd-workflows/src/generate_log.rs
mattklein123: OK will just fix this now.

PR #109, opened 2025-03-07
report crash reports with session of previous process

snowp:
For crash handling we'll assume that any of the newly reported crashes are associated with the previous process run. This seems sane since we clean up all the crash reports on each start immediately, thus we'd have previously cleaned up any older reports. Since we only check for reports on startup this also means that the current session cannot have emitted this. 

To achieve the correct ordering, we introduce the idea of "initial logs" that are passed to the pre-config buffer that will be replayed after the initial config log but before any of the otherwise queued up logs. This ensures that the crash logs are the first logs processed during app start if they exist, allowing them to be associated with the correct session/workflow engine instance. 

Fixes BIT-4827

### DIFF
--- a/bd-logger/src/builder.rs
+++ b/bd-logger/src/builder.rs
@@ -12,19 +12,20 @@ use crate::internal::InternalLogger;
 use crate::log_replay::LoggerReplay;
 use crate::logger::{ChannelPair, Logger};
 use crate::logging_state::UninitializedLoggingContext;
-use crate::{InitParams, LoggerHandle};
+use crate::InitParams;
 use bd_api::api::SimpleNetworkQualityProvider;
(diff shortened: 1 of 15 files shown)


bd-logger/src/builder.rs
```diff
@@ -254,12 +252,32 @@ impl LoggerBuilder {
 
     bd_client_common::error::UnexpectedErrorHandler::register_stats(&scope);
 
-    let logger_future = async {
+    let session_strategy = self.params.session_strategy;
+
+    let logger_future = async move {
+      // This is guaranteed to run before the initial configuration load, which means that
```
mattklein123, line 258: ?

mattklein123 (approved):

PR #110, opened 2025-03-07
allow creating a StreamHandle from a raw stream_id and server handle

snowp:
In the platform tests we marshal the stream as an integer into platform code, this should allow us to convert it back to a StreamHandle

### DIFF
--- a/bd-test-helpers/src/test_api_server.rs
+++ b/bd-test-helpers/src/test_api_server.rs
@@ -929,6 +929,15 @@ pub struct StreamHandle {
 
 impl StreamHandle {
   #[must_use]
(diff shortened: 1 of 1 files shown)


FranAguilera (approved):

bd-test-helpers/src/test_api_server.rs
```diff
@@ -928,6 +928,15 @@ pub struct StreamHandle {
 }
 
 impl StreamHandle {
+  #[must_use]
+  pub fn from_stream_id(stream_id: i32, server: &ServerHandle) -> Self {
```
FranAguilera, line 932: out of curiosity, where is this used?

snowp: Will push some code to the capture-sdk branch that uses it

PR #113, opened 2025-03-07
workflows: implement generate log action

mattklein123:
There are some other changes in here that are partially related.

1) I remove the enable workflows runtime key. I didn't realize how big this PR was going to get, so if we want to strip this out I can remove this part, though a little annoying.
2) In order to have stable integration tests I redid how client stat uploads work to make it fully deterministic.


bd-test-helpers/src/workflow.rs
```diff
@@ -664,3 +704,96 @@ pub fn make_tags(labels: BTreeMap<String, String>) -> LogFields {
...
+      })
+      .collect(),
+    ..Default::default()
+  }
+}
+
+#[must_use]
+pub fn make_generate_log_action_proto(
+  message: &'static str,
+  fields: &'static [(&'static str, TestFieldType)],
```
snowp, line 796: i would think a bunch of these don't have to specify 'static?

bd-logger/src/test/logger_integration.rs
```diff
@@ -926,11 +918,104 @@ fn workflow_flush_buffers_action_emits_synthetic_log_and_uploads_buffer_and_star
   });
 }
 
+#[test]
+fn workflow_generate_log_to_histogram() {
```
snowp, line 922: for this or a different test it would be nice to verify that we are setting session ID, timestamp, etc as expected on the generated log

bd-test-helpers/src/workflow.rs
mattklein123: It just avoids having to have a lifetime on the enum and then having that go everywhere even if anonymous...

Reflejo (approved):

PR #114, opened 2025-03-11
support inferring crash details based on crash report

snowp:
In order to support different crash formats on different platforms, this implements a fairly generic JSON extraction logic that can be used in order to pull out the crash reason from the various formats based on dynamic config


bd-crash-handler/src/lib.rs
```diff
@@ -92,10 +106,62 @@ impl Monitor {
...
+  fn guess_crash_details(
+    report: &[u8],
+    candidate_reason_paths: &[JsonPath],
+    candidate_details_path: &[JsonPath],
+  ) -> (Option<String>, Option<String>) {
+    // The report may come in either as a single JSON object or as a series of JSON objects.
+    // Defensively handle both to produce a list of objects that we want to inspect to infer the
+    // crash reason.
+    if report.first() != Some(&b'{') {
+      return (None, None);
```
FranAguilera, line 118: Wondering if instead of returning `None` would be good to indicate with a new type (e.g. MalformedContent) that this was unexpected file content and we couldn't guess the crash details 

Also, just a nit for now and not blocking on this but are we locking here to Json format? if client sent a raw stack trace, we could still have logic here (depending on platform) to retrieve the most relevant crash reason/title (

snowp: We're only handling JSON-like formats at the moment, but that doesn't stop us from adding support for other formats in the future depending on what else we want to support here

There is a bunch of error handling here that I'm skipping, we can talk more in the meeting today about how we want this stuff to work in general

FranAguilera: that's true, we can do further tweaks later on

FranAguilera (approved):

PR #117, opened 2025-03-13
extract timestamp from start of crash artifact

snowp:
Attempts to look for {timestamp}_ at the start of the crash file,
emitting this as the timestamp if parsing succeeded

### DIFF
--- a/bd-crash-handler/src/lib.rs
+++ b/bd-crash-handler/src/lib.rs
@@ -13,6 +13,7 @@ use bd_log_primitives::{AnnotatedLogField, LogFields};
 use bd_runtime::runtime::{ConfigLoader, StringWatch};
 use bd_shutdown::ComponentShutdown;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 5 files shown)


FranAguilera (approved):

bd-crash-handler/src/lib.rs
```diff
@@ -265,6 +267,20 @@ impl Monitor {
...
         };
+
+        let timestamp = path
+          .file_name()
+          .and_then(|name| name.to_str())
+          .and_then(|name| {
+            name.split('_').next().and_then(|timestamp| {
+              let Ok(timestamp) = timestamp.parse() else {
+                log::debug!("Failed to parse timestamp from file name: {:?}", name);
+                return None;
```
murki, line 278: should we return Now as a fallback?

snowp: I do already elsewhere but I can move the handling here since I didn't end up having to test it at this level

PR #118, opened 2025-03-13
track ootb fields as global state to provide information to crash handler

snowp:
In order to produce the fields we want on the crash log that is emitted on next app launch we'll track the last seen ootb field provider values in the persistent storage. This allows us to recover this at the next app launch and include the last observed fields on the crash log


bd-logger/src/global_state.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use bd_device::Store;
+use bd_key_value::{Key, Storable};
+use bd_log_primitives::{LogField, LogFields};
+use std::sync::Arc;
+
+const KEY: Key<State> = Key::new("global_state");
+
+#[derive(Debug, serde::Serialize, serde::Deserialize, Default, PartialEq, Eq)]
+struct State(Vec<(String, String)>);
```
mattklein123, line 20: This is going to get checked a lot, I would probably switch this to a HashMap for quicker lookup in the common case since we can't guarantee the order of the incoming.

mattklein123 (approved):

bd-logger/src/global_state.rs
snowp: I had that originally but changed it back to this since I realized that we haven't deduped logs at this point and I didn't want to have to construct a new HashMap just to make the comparison. Since the metadata providers are all evaluated in order it seems like we can rely on the order being consistent, but maybe this is a good time to switch over to using a map type everywhere?

snowp: Of course that was a ton of work, probably worth another cursory review

:w
```diff
@@ -0,0 +1,231 @@
+// shared-core - bitdrift's common client/server libraries
```
mattklein123, line 1: vim issue? :)

bd-log-primitives/src/lib.rs
```diff
@@ -112,67 +114,48 @@ pub type LogFieldValue = StringOrBytes<String, Vec<u8>>;
 //
 
 /// The list of log fields annotated with extra information.
-pub type AnnotatedLogFields = Vec<AnnotatedLogField>;
+pub type AnnotatedLogFields = HashMap<String, AnnotatedLogField>;
```
mattklein123, line 117: minor item that we can do as a follow up. We should see if we are already bringing any of the more performant hashers vs. the one the stdlib uses by default which is HashDoS resistant which we don't need. It would be nice to speed this up with a more performant hasher. Since we are already using a type alias for this we can always swap that in later, might as put a TODO in.

bd-crash-handler/src/lib.rs
```diff
@@ -291,21 +291,20 @@ impl Monitor {
         // TODO(snowp): Eventually we'll want to upload the report out of band, but for now just
         // chuck it into a log line.
         logs.push(CrashLog {
-          fields: vec![
-            AnnotatedLogField::new_ootb("_crash_artifact".into(), contents.into()).into(),
-            AnnotatedLogField::new_ootb(
+          fields: [
+            ("_crash_artifact".into(), contents.into()),
```
mattklein123, line 295: Another follow up that I think we should do at this point is that there are many places where the key is just a static string. I think based on how you have it setup you could pretty easily have a FieldKey type which operates on &'static string. You could have some of these be lazy init Arc but that seems more error prone vs. just supporting it directly. There is no reason to copy these things over and over, especially for the higher rate ones.

mattklein123 (commented):
Nice I think this is a great improvement. Some comments, feel free to add as TODOs and we can consider them in follow ups.

bd-log-primitives/src/lib.rs
snowp: Switched out to ahash and it didn't blow up the binary size, so that seems like a no-brainer 

Binary size does seem to have creeped up so maybe I will look at that separately

bd-crash-handler/src/lib.rs
snowp: I just switched FIeldKey to be a Cow which was relatively painless, seems like a nice improvement 

We could potentially also support Arc<String> but we wouldn't benefit from that in many places so maybe not worth the complexity. Using a symbol store might be helpful to avoid re-creating the same keys over and over again but that's a bigger project

PR #123, opened 2025-03-21
Allow override occurred_at without a session ID

Reflejo:


snowp (approved):
Would be nice with a new test case for this flow but seems OK to me

Another odd thing is that by placing the log in the past it's possible to put it at a point in time where a previous session was active which might be a bit strange but shouldn't break anything

PR #139, opened 2025-04-22
make log emitted by crash handler conform to AppExit

snowp:


bd-crash-handler/src/lib.rs
```diff
@@ -304,11 +304,12 @@ impl Monitor {
           [
             ("_crash_artifact".into(), contents.into()),
             (
-              "_crash_reason".into(),
+              "_app_exit_info".into(),
               crash_reason.unwrap_or_else(|| "unknown".to_string()).into(),
             ),
+            ("_app_exit_source".into(), "bitdrift".into()),
```
murki, line 310: can we distinguish between vendor and built-in? I think that should be the key distinction

murki (approved):

bd-crash-handler/src/lib.rs
FranAguilera: +1

FranAguilera:
This is being handled in here https://github.com/bitdriftlabs/shared-core/pull/155

PR #140, opened 2025-04-23
add report format converter util

kattrali:
Adds a runnable tool for converting reports between JSON and binary format. Runs with:

```sh
# prints to stdout
$ cargo run bd-report-convert to-json path/to/report.bin
```

```sh
# writes to a file
$ cargo run bd-report-convert to-bin path/to/report.json report.bin
```
the latter one requiring an output file path to avoid accidentally spewing binary bits into a terminal (something I have definitely never done).


bd-report-convert/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+use std::{env, fs, slice};
+
+extern "C" {
+  fn bdrr_make_bin_from_json(json_data_path: *const c_char, length_or_err: *mut i32) -> *const u8;
+
+  fn bdrr_print_json(bin_data_path: *const c_char) -> i32;
+}
+
+fn print_usage(program: &str) -> ExitCode {
+  println!("converter for the bd issue report format\n");
```
snowp, line 20: `eprintln` for all these error messages?

bd-report-convert/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+  println!("usage: {program} to-json REPORT_BIN_PATH");
+  println!(
+    "       {} to-bin  REPORT_JSON_PATH DEST_PATH",
+    " ".repeat(program.len())
+  );
+  ExitCode::from(64) // EX_USAGE error
+}
+
+pub fn main() -> Result<ExitCode, std::io::Error> {
+  let args: Vec<String> = env::args().collect();
```
snowp, line 30: you could use `clap` to manage cli arugments but this is simple enough so up to you

bd-report-convert/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use std::ffi::{c_char, CString};
+use std::io::Write;
+use std::process::ExitCode;
+use std::{env, fs, slice};
+
+extern "C" {
+  fn bdrr_make_bin_from_json(json_data_path: *const c_char, length_or_err: *mut i32) -> *const u8;
+
+  fn bdrr_print_json(bin_data_path: *const c_char) -> i32;
```
snowp, line 16: just curious what does bdrr stand for

bd-report-convert/src/glue.cpp
```diff
@@ -0,0 +1,102 @@
...
+ * @param json_data_path  the path to the data to convert in JSON format
+ * @param length_or_err   destination for the length of the buffer or an error
+ *
+ * @return an allocated buffer containing the binary data or null if parsing
+ *         failed and the value of length_or_err should be checked for the
+ *         problem
+ */
+uint8_t *bdrr_make_bin_from_json(
+    char *json_data_path,
+    int32_t */* out */length_or_err
```
snowp, line 65: GH highlighting really didn't like where this comment is lol

bd-report-convert/src/glue.cpp
```diff
@@ -0,0 +1,102 @@
...
+enum BDReaderErr {
+  BDReaderErrCouldNotReadSpec = -1,
+  BDReaderErrCouldNotReadInputFile = -2,
+  BDReaderErrCouldNotParseSpec = -3,
+  BDReaderErrCouldNotParseInputData = -4,
+  BDReaderErrCouldNotValidateOutput = -5,
+};
+
+#define SPECDIR "api/src/bitdrift_public/fbs/issue-reporting/v1"
+const char *SPECPATH = SPECDIR "/report.fbs";
```
snowp, line 29: maybe just inline the define here? doesn't seem like it's adding much

bd-report-convert/src/glue.cpp
```diff
@@ -0,0 +1,102 @@
...
+ *         problem
+ */
+uint8_t *bdrr_make_bin_from_json(
+    char *json_data_path,
+    int32_t */* out */length_or_err
+) {
+  std::string json_data;
+  std::string schema_data;
+
+  if (!flatbuffers::LoadFile(SPECPATH, false, &schema_data)) {
```
snowp, line 70: Does this mean that we also need to have this available at runtime? probably OK as long as we're expecting people to run via cargo run but I imagine this would break otherwise? We could embed the file into the binary without too much trouble but not sure if it matters

ci/setup.sh
```diff
@@ -9,5 +9,9 @@ git submodule update --init --recursive
 # Install library compile deps.
 if ! [[ -z "$RUNNER_TEMP" ]]; then
   sudo apt-get update
-  sudo apt-get install lld
+  sudo apt-get install lld cmake
 fi
+
+# Build thirdparty deps
+cmake -Bthirdparty/flatbuffers/build thirdparty/flatbuffers
+cmake --build thirdparty/flatbuffers/build --target flatbuffers
```
snowp, line 17: would be nice to have this in a build.rs that picked up changes to that dir but probably fine for now

snowp (commented):
Obviously would be nice for this to just use rust but I'm guessing that the rust library doesn't have the same JSON support?

PR #141, opened 2025-05-02
add module for writing a report via ffi

kattrali:
Depends on #142 

This is a thin(ish) wrapper around the rust interface to report generation to make something suitable for C FFI. The main chunk of logic is `bd-report-writer/src/ffi.rs`, and much of the rest is ~generated code and~ tests. It looks like a lot but is less than it seems.

* includes the rust binding to the api spec - this is the avenue by which we'll have it in the SDK (for now?) and can replace the copy in lapi (once this stabilizes)
* generates a header for FFI via `cbindgen`
* the secondary package `bd-report-writer-tests` is an accomodation for the fact that test binaries can't have a separate build script (so I can exercise the C interface without bundling test functions into the library itself). Maybe there's a better way to package this?


bd-report-writer/src/ffi.rs
```diff
@@ -0,0 +1,569 @@
...
+
+/// Entry point to creating a new Report
+///
+/// * `report_type` - raw value of the `ReportType` which will be written
+/// * `sdk_id`      - identifier of the integration sending the report
+/// * `sdk_version` - version of the integration sending the report
+///
+/// # Safety
+/// This function returns an object which is not managed by normal lifetime
+/// semantics and must be discarded using `bdrw_dispose_buffer_handle()`
```
snowp, line 154: i think pedantically this would point out the expectations of the parameters that make this function safe to call, e.g. that all the pointers are dereferncable and that `handle` can be written to

not sure if you want to care about putting all the unsafe assumptions in the safety block or just make reasonable assumptions about usage

bd-report-writer/src/ffi.rs
```diff
@@ -0,0 +1,569 @@
...
+#[unsafe(no_mangle)]
+extern "C-unwind" fn bdrw_create_buffer_handle(
+  handle: BDProcessorHandle,
```
snowp, line 157: missing from docs

bd-report-writer/src/ffi.rs
```diff
@@ -0,0 +1,569 @@
...
+  errors: Vec<WIPOffset<Error<'a>>>,
+  report_type: ReportType,
+  sdk: WIPOffset<SDKInfo<'a>>,
+  app: Option<WIPOffset<AppMetrics<'a>>>,
+  device: Option<WIPOffset<DeviceMetrics<'a>>>,
+}
+
+// Type alias for the gross pointers passed around for FFI to improve
+// general readability
+pub type BDProcessorHandle = *mut *const c_void;
```
snowp, line 36: you could do a newtype with repr(transparent) if you want to pretend like you have more type safety

bd-report-writer/src/ffi.rs
```diff
@@ -0,0 +1,569 @@
...
+///
+/// # Safety
+/// All actions involving the memory buffer returned from
+/// `bdrw_get_completed_buffer()` should be completed prior to calling this
+/// function.
+#[unsafe(no_mangle)]
+extern "C" fn bdrw_dispose_buffer_handle(handle: BDProcessorHandle) {
+  unsafe {
+    drop(Box::<ReportProcessor<'_>>::from_raw(*handle as *mut _));
+    *handle = null();
```
snowp, line 245: you could check for null at the start here for same basic defensive programming

bd-report-writer/src/ffi.rs
```diff
@@ -0,0 +1,569 @@
...
+          path,
+          load_address: image.load_address,
+        },
+      ))
+    } else {
+      None
+    }
+  }) {
+    processor.binary_images.push(unpacked_image);
+    pack_processor!(handle, processor);
```
snowp, line 275: instead of pack/unpack you could just get a &mut handle to the processor, you're already assuming exclusive access so it could clean it up a little bit

bd-report-writer/src/ffi.rs
snowp: we kinda do this with with LoggerId in platform/shared if you want to compare notes

bd-report-writer/src/generated/report_generated.rs
snowp, line 1: in my branch I put this in bd-proto next to the buffer_log.fbs that we already generate flatbuffer bindings for, seems like a reasonable place to put it

bd-report-writer/src/ffi.rs
```diff
@@ -0,0 +1,569 @@
...
+    },
+  );
+  processor.errors.push(error);
+  pack_processor!(handle, processor);
+  true
+}
+
+/// Add device info to the report
+#[unsafe(no_mangle)]
+extern "C" fn bdrw_add_device(
```
snowp, line 353: any reason to not put C-unwind on all of these?

snowp (approved):

PR #144, opened 2025-05-07
rework client metrics to use new API

mattklein123:
1) Workflow metrics are now strongly typed by ID.
2) Cardinality limits are now per action ID.
3) Correctly respect limits when merging metrics to disk.

### DIFF
--- a/bd-runtime/src/runtime.rs
+++ b/bd-runtime/src/runtime.rs
@@ -868,8 +868,9 @@ pub mod stats {
     5.minutes()
   );
 
-  // This controls how many unique counters we allow before rejecting new metrics. This limit
-  // prevents unbounded growth of metrics, which could result in the system running out of memory.
+  // This controls how much tag cardinality we allow before rejecting new metrics *per workflow*.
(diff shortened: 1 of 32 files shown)


bd-runtime/src/runtime.rs
```diff
@@ -868,8 +868,9 @@ pub mod stats {
     5.minutes()
   );
 
-  // This controls how many unique counters we allow before rejecting new metrics. This limit
-  // prevents unbounded growth of metrics, which could result in the system running out of memory.
+  // This controls how much tag cardinality we allow before rejecting new metrics *per workflow*.
+  // This limit prevents unbounded growth of metrics, which could result in the system running out
+  // of memory.
```
mattklein123, line 873: Do we still want a global limit? Not sure if it's worth it.

snowp (approved):

bd-runtime/src/runtime.rs
snowp: doesn't seem super important

PR #145, opened 2025-05-08
feat(crash-handler): infer crash details from bin format files

kattrali:
unsure how practical this approach is (pulling in fbs to check random files in the directory?) but this ensures bin files have a reason and details for logging


bd-crash-handler/src/lib.rs
```diff
@@ -122,6 +125,17 @@ impl Monitor {
     candidate_reason_paths: &[JsonPath],
     candidate_details_path: &[JsonPath],
   ) -> (Option<String>, Option<String>) {
+    if let Ok(bin_report) = root_as_report_with_opts(&VerifierOptions::default(), report) {
```
snowp, line 128: I think `root_as_report` is the same without having to pass VerfierOptions

snowp (approved):
Seems reasonable to me. We can take this and check how much the binary size imapct of this is 

I kinda wonder if the way flatbuffers work we can LTO away all the code related to fields that we are not referencing in code since there isn't a monster parse function like protobuf 

I'll move the code I added for making reports in test and we can add a simple test case for this

PR #146, opened 2025-05-12
workflows: bump state file name

mattklein123:
When we added compression/CRC to the state file, this causes a crash if the library is downgraded as bincode is not very resilient. We will look at no longer using bincode in a future version.


FranAguilera:
Verified with a downgrade to 0.17.11

FranAguilera (approved):

PR #152, opened 2025-05-16
implement out of brand report uploads

snowp:
Implements a uploader that orchestrates intents and uploads for mostly opaque artifacts that can be uploaded out of band of the typical log uploads. 

Switches over the crash report uploading flow to use the out of band upload mechanism instead of the previous mechanism which relies on embedding the report in a log field


bd-crash-handler/src/lib.rs
```diff
@@ -311,30 +315,32 @@ impl Monitor {
...
-          ]
-          .into_iter(),
-        );
-
-        // TODO(snowp): For now everything in here is a crash, eventually we'll need to be able to
-        // differentiate.
-        // TODO(snowp): Eventually we'll want to upload the report out of band, but for now just
-        // chuck it into a log line.
-        logs.push(CrashLog { fields, timestamp });
+        if let Ok(artifact_id) = self.artifact_client.enqueue_upload(contents).await {
```
mattklein123, line 318: I don't have a good answer for this, but if this upload eventually gets intent rejected for whatever reason we are going to have to know about that somehow on the backend. So I guess depending on the use case we may have to track rejected state (like if it's duplicated/collapsed or whatever).

bd-crash-handler/src/uploader.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+pub struct UploadClient {
+  upload_tx: tokio::sync::mpsc::Sender<NewUpload>,
+}
+
+#[async_trait]
+impl Client for UploadClient {
+  /// Dispatches a payload to be uploaded, returning the associated artifact UUID.
+  async fn enqueue_upload(&self, contents: Vec<u8>) -> anyhow::Result<Uuid> {
```
mattklein123, line 42: nit: async not needed if we are doing try_send which makes sense to avoid overflow/blocking.

bd-crash-handler/src/uploader.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+impl Uploader {
+  pub fn new(
+    file_system: Arc<dyn FileSystem>,
+    data_upload_tx: tokio::sync::mpsc::Sender<DataUpload>,
+    time_provider: Arc<dyn TimeProvider>,
+    shutdown: ComponentShutdown,
+  ) -> (Self, UploadClient) {
+    let (upload_tx, upload_rx) = tokio::sync::mpsc::channel(2);
```
mattklein123, line 73: Does 2 have special meaning? We probably do this elsewhere but I can't remember why. Like vs 1 or whatever.

bd-crash-handler/src/uploader.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    (uploader, client)
+  }
+
+  pub async fn run(mut self) -> anyhow::Result<()> {
+    self.initialize().await?;
+
+    // TODO(snowp): Due to the way this is structured we only pop values off the queue of inbound
+    // requests while we have no pending uploads. It may need to be two tasks that interact with
+    // the index through a mutex.
```
mattklein123, line 94: FWIW I think this is required. During no network scenarios we will start dropping right away. We should keep accepting uploads up to the limits imposed in the other TODOs.

bd-crash-handler/src/uploader.rs
```diff
@@ -0,0 +1,282 @@
...
+      name: uuid.clone(),
+      time: self.time_provider.now().into_proto(),
+      pending_intent_negotation: true,
+      ..Default::default()
+    });
+
+    self
+      .file_system
+      .write_file(&REPORT_DIRECTORY.join(uuid), &contents)
+      .await?;
```
mattklein123, line 198: I think at this point every file we write has to have at least CRC protection. We don't necessarily have to use the compression approach (that is probably overkill and perhaps we should redo some of that to just write the file + a CRC at the end), but I think we have to make sure that it is in place. So either this code has to do it or we have to assume the calling code is going to do it. Then we have to decide whether it will be checked on read back or we will assume the server is going to check it. We should discuss this offline with the team to figure out what we want to do here.

mattklein123 (commented):
LGTM at a high level modulo all the TODOs and some comments.

bd-crash-handler/src/uploader.rs
snowp: 2 was just something I was playing with to get some tests to pass. The current producer code can emit multiple reports on startup (unclear if we need this) and so having this be 1 means that it's possible that it immediately overflows and we drop the report. 

We could use 1 here, but for cases where we are using try_send this is what controls how much we can buffer while we wait for the consumer, so setting an appropriate value here is more important than in other cases where we just cause back-pressure on the channel being full. In async_log_buffer we set the channel size to 512 for example

We probably don't need to support multiple this for reports at all so we can maybe just start with channel(1) and worry about the buffer capacity when/if we start having other things produce artifacts to be uploaded

mattklein123: I wasn't implying that we should use 1, I was just trying to understand where 2 comes from. Whatever we end up with I would just add some comments.

bd-runtime/src/runtime.rs
```diff
@@ -806,3 +806,19 @@ pub mod platform_events {
...
 }
+
+pub mod artifact_upload {
+  bool_feature_flag!(Enabled, "artifact_upload.enabled", false);
+
+  int_feature_flag!(MaxPendingEntries, "artifact_upload.max_pending_entries", 10);
+
+  int_feature_flag!(BufferCountLimit, "artifact_upload.buffer_count_limit", 100);
+
+  static ONE_MEGABYTE: u32 = 1 << 20;
```
mattklein123, line 817: super nit: personally I find `1024 * 1024` easier to spot check but up to you.

bd-artifact-upload/src/uploader.rs
```diff
@@ -0,0 +1,666 @@
...
+    let uuid = uuid::Uuid::new_v4();
+
+    let result = self
+      .upload_tx
+      .try_send(NewUpload {
+        uuid,
+        contents,
+        state,
+      })
+      .inspect_err(|e| log::warn!("failed to enqueue artifact upload: {e:?}"));
```
mattklein123, line 129: Related to my other comment, should this also increment the dropped metric? That would cover it.

bd-artifact-upload/src/uploader.rs
```diff
@@ -0,0 +1,666 @@
...
+    Ok(uuid)
+  }
+}
+
+#[derive(thiserror::Error, Debug)]
+enum Error {
+  #[error("Task is shutting down")]
+  Shutdown,
+  #[error("Unhandled error: $1")]
+  Unhandled(anyhow::Error),
```
mattklein123, line 143: low pri: we could move the map_error! thing from internal_core to here.

bd-artifact-upload/src/uploader.rs
```diff
@@ -0,0 +1,666 @@
...
+    .artifact
+    .into_iter()
+    .collect();
+
+    // Ensure that the files stored on disk pending upload and the index are in sync. If either the
+    // file is missing for an index entry or a file exists without an index entry they can never
+    // be uploaded, so just clean them up.
+
+    // TODO(snowp): Should we check for crc integrity at this point? Currently we only do so when
+    // we are considering a file for upload.
```
mattklein123, line 368: I think it's probably ok to do when you upload but no strong opinion.

bd-artifact-upload/src/uploader.rs
```diff
@@ -0,0 +1,666 @@
...
+    }
+    if let Some(task) = self.intent_task_handle.take() {
+      task.abort();
+    }
+  }
+
+  async fn track_new_upload(&mut self, uuid: Uuid, contents: Vec<u8>, state: LogFields) {
+    // If we've reached our limit of entries, stop the entry currently being uploaded (the oldest
+    // one) to make space for the newer one.
+    if self.index.len() == usize::try_from(*self.max_entries.read()).unwrap_or_default() {
```
mattklein123, line 482: I think like the bounded buffer we should probably do this both by count and size just given that these files could be large? I would just add a todo/ticket about this.

bd-artifact-upload/src/uploader.rs
```diff
@@ -0,0 +1,666 @@
...
+
+      self.stats.dropped.inc();
+      self.stop_current_upload();
+      self.index.pop_front();
+    }
+
+    let uuid = uuid.to_string();
+
+    // Add a CRC trailer to ensure we don't try to upload a malformed report.
+    let contents = write_checksummed_data(&contents);
```
mattklein123, line 493: Per your other point about passing a file around, we could avoid another full copy of this memory here since we could stream it all from the file.

bd-artifact-upload/src/uploader.rs
```diff
@@ -0,0 +1,666 @@
...
+  async fn upload_artifact(
+    data_upload_tx: tokio::sync::mpsc::Sender<DataUpload>,
+    contents: Vec<u8>,
+    name: String,
+  ) -> Result<()> {
+    let path = REPORT_DIRECTORY.join(&name);
+    log::debug!("uploading artifact: {}", path.display());
+
+    // Use exponential backoff to avoid retrying over and over again in case something is going
+    // wrong. We put no overall tiemout as the device might be offline for a long time and we want
```
mattklein123, line 581:
```suggestion
    // wrong. We put no overall timeout as the device might be offline for a long time and we want
```

bd-artifact-upload/src/uploader.rs
```diff
@@ -0,0 +1,666 @@
...
+      let delay = retry_backoff.next_backoff().unwrap();
+      log::debug!("intent negotiation for artifact: {id} failed, retrying in {delay:?}");
+      tokio::time::sleep(delay).await;
+    }
+  }
+
+  fn make_retry_policy() -> ExponentialBackoff<SystemClock> {
+    // TODO(snowp): Make more of these parameters runtime configurable.
+
+    // This default sets an initial timeout of 500s with a max timeout of 1 min (jittered).
```
mattklein123, line 657:
```suggestion
    // This default sets an initial timeout of 500ms with a max timeout of 1 min (jittered).
```

I assume? For this one can we just share whatever backoff policy we are using in API? I think I'm using much longer settings there but I can't remember. We can move that backoff policy to client-common?

mattklein123 (approved):
I reviewed to the best of my ability. We will just need to do a lot of testing. Looks awesome!

bd-artifact-upload/src/uploader.rs
snowp: The bounded buffer comes with a bunch of stats already so I was planning on using that, including whether the drop was due to size or count

bd-artifact-upload/src/uploader.rs
snowp: The API one seems to use 30s/30min but has their own runtime flags that drive them - I wonder if we should consolidate the runtime flags and just have one configuration for general backoff or if it makes sense to keep them separate?

mattklein123: My vote would be to just consolidate.

PR #154, opened 2025-05-20
cache config safety part 2

mattklein123:
1) Mark config safe if the server reports the relevant config
   is up to date.
2) Mark config safe if we have been up for long enough to consider
   the device offline. This is currently 15s.
3) Simplify the retry_count file to be a single byte. 
4) Break circular deps and various cleanups to make testing simpler. 
5) (Unrelated) Remove the compression feature flag which we will never flip.


bd-api/src/api.rs
```diff
@@ -676,14 +642,24 @@ impl Api {
...
           stream_state.initialize_stream_settings(stream_settings);
 
           self
             .handle_responses(remaining_responses, &mut stream_state)
             .await?;
+
+          // Let the all the configuration pipelines know that we are able to connect to the
+          // backend. We do this after we process responses that immediately followed the handshake
+          // in order to make it possible for the server to push down configuration that
+          // would get applied as part of the handshake response.
```
snowp, line 657: Is the bit about processing the responses after the handshake still relevant?

snowp (approved):

bd-api/src/api.rs
mattklein123: No let me fix that comment, thanks for catching.

PR #165, opened 2025-06-05
session strategy: fix potential deadlock

mattklein123:
There are cases in which the session_id() function can get called from within the change callback so make sure this works.


murki (approved):
thx

PR #166, opened 2025-06-06
implement initial sleep mode support

mattklein123:
During sleep mode the stats upload interval is greatly reduced. Everything else is unchanged currently. Analytics about the mode are sent at handshake start as well as within each ping.

Fixes BIT-5059

### DIFF
--- a/bd-logger/src/test/logger_integration.rs
+++ b/bd-logger/src/test/logger_integration.rs
@@ -79,6 +79,31 @@ use time::macros::datetime;
 use time::OffsetDateTime;
 
 #[test]
+fn sleep_mode() {
+  let mut setup = Setup::new_with_options(SetupOptions {
+    start_in_sleep_mode: true,
+    ..Default::default()
+  });
+
+  setup.logger_handle.transition_sleep_mode(false);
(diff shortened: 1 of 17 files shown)


bd-logger/src/test/logger_integration.rs
```diff
@@ -1767,24 +1792,8 @@ fn stats_upload() {
...
-    Some(2),
-  );
-  assert_eq!(
-    stat_upload.get_counter("logger:logs_received", labels! {"log_level" => "warn"}),
-    Some(1),
-  );
-  assert_eq!(
-    stat_upload.get_counter("logger:logs_received", labels! {"log_level" => "error"}),
-    Some(2),
+    stat_upload.get_counter("logger:logs_received", labels! {}),
```
snowp, line 1795: not sure we need this one either? logs by log levels is even a instant insights default workflow

snowp (approved):

bd-logger/src/test/logger_integration.rs
mattklein123: OK I figured it might be useful just to see what is coming in. I can remove this also.

PR #185, opened 2025-06-30
workflows: cleanup 2 runtime flags

mattklein123:
1) "workflows.state_periodic_write_interval_ms": This flag is redundant
   as we do periodic persistence processing on every log processed and
   logs are received frequently.
2) "workflows.traversals_global_count_limit": With the removal of
   parallel execution mode this flag doesn't provide a lot of value and
   removing it cleans up a lot of complexity.

Also clean up some metrics that I don't think provide any real operational value.


mattklein123:
> I think some of the metrics might have had some test value to make sure that run management was done as expected, not sure if this costs us some test assertion value. Would be a painful audit process to tell for sure, might not be worth it

Yeah I thought about this. I agree it would be incredibly painful. I think in many cases it is covered by looking at the active workflow states, but possibly not 100%. I can go back and try to check if there are places that don't have further active state checks and try to add them.

PR #196, opened 2025-07-18
fix(crash-handler): load state fields from reports

kattrali:
load changeable fields like timestamp, OS version, and app version from report contents where available, which should reduce the possibility of old reports on disk causing the appearance of errors in new app versions.

part of BIT-5724

### DIFF
--- a/bd-crash-handler/src/lib.rs
+++ b/bd-crash-handler/src/lib.rs
@@ -15,6 +15,7 @@ mod json_extractor;
 use anyhow::bail;
 use bd_log_primitives::{LogFieldKey, LogFieldValue, LogFields, LogMessageValue};
 use bd_proto::flatbuffers::report::bitdrift_public::fbs;
+use bd_proto::flatbuffers::report::bitdrift_public::fbs::issue_reporting::v_1::{Platform, Report};
 use bd_runtime::runtime::{BoolWatch, ConfigLoader, StringWatch};
 use bd_shutdown::ComponentShutdown;
 use fbs::issue_reporting::v_1::root_as_report;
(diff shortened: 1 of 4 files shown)


bd-crash-handler/src/lib.rs
```diff
@@ -150,38 +151,89 @@ impl Monitor {
...
+    ]);
+
+    if let Some(app_id) = app_id {
+      fields.extend([("app_id".into(), app_id.into())]);
+    }
+
+    if platform == Platform::Android {
+      let version_code =
+        build_number.map_or_else(|| "unknown".to_string(), |b| b.version_code().to_string());
+      fields.extend([("_app_version_code".into(), version_code.into())]);
```
snowp, line 199: mentioned this in the other PR but is there no ios equivalent to this?

bd-crash-handler/src/lib.rs
```diff
@@ -150,38 +151,89 @@ impl Monitor {
...
+    }
+    (timestamp, fields)
+  }
+
+  fn guess_crash_details<'a>(
```
snowp, line 204: nit but this function name seems off now that it also returns the report

bd-crash-handler/src/monitor_test.rs
```diff
@@ -309,6 +319,97 @@ async fn config_file() {
...
+    &mut builder,
+    &ErrorArgs {
+      name,
+      reason,
+      stack_trace: None,
+      relation_to_next: ErrorRelation::CausedBy,
+    },
+  );
+  let errors = Some(builder.create_vector::<WIPOffset<Error<'_>>>(&[error]));
+  let app_id = Some(builder.create_string("com.example.foo.widget"));
```
snowp, line 337: I have some code internally that lets us replace this with 
```
let app_id = "com.example.foo.widget".to_fb(&mut builder)
```
We could bring this into this repo if we think it helps readability

bd-crash-handler/src/monitor_test.rs
```diff
@@ -309,6 +319,97 @@ async fn config_file() {
...
+
+  let uuid = "12345678-1234-5678-1234-5678123456aa".parse().unwrap();
+  setup.expect_artifact_upload(
+    data,
+    uuid,
+    [
+      ("other_stuff".into(), "foo".into()),
+      ("app_version".into(), "4.15".into()),
+      ("os_version".into(), "3.1".into()),
+      ("app_id".into(), "com.example.foo.widget".into()),
```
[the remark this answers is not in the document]
snowp: When it comes to the app ID associated with the crash for the purpose of filtering etc I think we might pin this to the one on the API handshake, so might not matter much what we do here

bd-crash-handler/src/monitor_test.rs
snowp: https://github.com/bitdriftlabs/shared-core/pull/198

snowp (approved):

PR #200, opened 2025-07-30
Cleanup field providers traits

snowp:
Previously there would be two `FieldProvider` traits that worked slightly differently:
- The bd-log-filter trait would expect log_level/type/message to be encoded as a field, so it would extract those based on the key
- The bd-matcher one would only operate on Fields except in one impl which we never seemed to rely on

With this change we change these to
- The bd-log-filter now is just a standalone function that explicitly operates differently than the other field interactions
- The bd-matcher one goes away and we operate directly on FieldRef instead of using a trait indirection. This adds some complexity in a few call sites but in general makes the code a bit easier to follow

As a bonus this breaks the dependency between bd-log-matcher and bd-matcher and should make a future cleanup there a bit more manageable


mattklein123 (approved):
Nice!

PR #201, opened 2025-07-31
implement handling of legacy matcher in bd-log-matcher

snowp:
Implements conversion from the legacy Matcher format into a bd-log-matcher Tree, allowing us to get rid of bd-matcher. 

Ideally we should stop sending the legacy format to newer clients but this requires a bit more careful rollout so starting with this that will let us remove a confusing crate


mattklein123 (approved):
Nice!!!

PR #203, opened 2025-08-01
move crash log processing into platform-level trigger

kattrali:
Moves crash log processing from immediate to into a function on `Logger` which will process them, once, on demand.

part of BIT-5823

### todo

- [x] refactor `CrashLog` into `LogLine` (?) which should make the sending process less redundant
- [x] cleanup in `AsyncLogBuffer` tests, which should have some actual coverage for the new receiver
- [x] assign crash monitor instance to logger in a less hamfisted way
- [x] fix the logger tests

### testing

- checked that processing did not cause a workflow state clean

verified contents, timestamp, and session ID for the following scenarios:

- crash followed by session ID change at next launch
- launch crash loop
- crash without session ID change at next launch

### DIFF
--- a/bd-logger/src/async_log_buffer.rs
+++ b/bd-logger/src/async_log_buffer.rs
@@ -11,7 +11,7 @@ mod async_log_buffer_test;
 
 use crate::device_id::DeviceIdInterceptor;
 use crate::log_replay::LogReplay;
-use crate::logger::with_thread_local_logger_guard;
+use crate::logger::{ReportProcessingRequest, with_thread_local_logger_guard};
 use crate::logging_state::{ConfigUpdate, LoggingState, UninitializedLoggingContext};
 use crate::metadata::MetadataCollector;
 use crate::network::{NetworkQualityInterceptor, SystemTimeProvider};
@@ -21,7 +21,7 @@ use anyhow::anyhow;
 use bd_bounded_buffer::{MemorySized, Receiver, Sender, TrySendError, channel};
 use bd_buffer::BuffersWithAck;
 use bd_client_common::error::{handle_unexpected, handle_unexpected_error_with_details};
-use bd_crash_handler::{Monitor, global_state};
+use bd_crash_handler::global_state;
 use bd_device::Store;
 use bd_log_metadata::MetadataProvider;
 use bd_log_primitives::{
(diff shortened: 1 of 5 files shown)


bd-logger/src/builder.rs
```diff
@@ -297,39 +299,12 @@ impl LoggerBuilder {
...
-            .unwrap_or_else(|| session_strategy.session_id()),
-          occurred_at: crash_log.timestamp,
-          // Always capture the session when we process a crash log.
-          // TODO(snowp): Ideally we should include information like the report and client side
-          // grouping here to help make smarter decisions during intent negotiation.
-          capture_session: Some("crash_handler".to_string()),
-        })
-        .collect();
-
       try_join!(
```
mattklein123, line 302: This is the top level future that basically never exits. I don't think this change and the related change in running the logger runtime is what you want. Basically just create the monitor above and return it, unless I'm missing something? then you would communicate with it using the channel.

bd-logger/src/logger.rs
```diff
@@ -534,6 +542,19 @@ impl Logger {
...
   }
 
+  pub fn process_crash_reports(&mut self, session_id: String) -> anyhow::Result<()> {
+    let Some(crash_monitor) = self.crash_monitor.take() else {
+      anyhow::bail!("no crash monitor set");
+    };
+
+    if let Some(tx) = self.report_processor_tx.take() {
+      tx.send((crash_monitor, session_id))
+        .map_err(|_| anyhow!("oneshot report processor channel closed"))
```
mattklein123, line 552: nit

```suggestion
      Ok(tx.send((crash_monitor, session_id))?)
```

bd-logger/src/async_log_buffer.rs
```diff
@@ -690,10 +689,45 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
           if let Some(pre_config_log_buffer) = maybe_pre_config_buffer {
             self.maybe_replay_pre_config_buffer_logs(
                 pre_config_log_buffer,
-                &mut initial_logs
+                &mut vec![]
             ).await;
           }
         },
+        Ok((crash_monitor, session_id)) = &mut self.report_processor_rx => {
```
mattklein123, line 696: I'm not 100% sure but I don't think a oneshot receiver is meant to be polled multiple times. I don't know if this will spin or do something else bad. I think you either need to put this into an option and then clear it when done, or fuse() it.

mattklein123 (changes requested):
Nice definitely on the right track.

mattklein123 (changes requested):

PR #206, opened 2025-08-13
add rudimentary logger runner

kattrali:
Adds a new bin command runnable with cargo: `cargo run --bin logger-cli`. 

Has a few commands:

* `log`: emits a log with the fields requested
* `enqueue-artifacts`: add files to the reports dir
* `upload-artifacts`: run `logger.process_new_reports()`, uploading artifacts and generating log lines
* `new-session`: regenerates the cached session ID

[sample session](https://timeline.bitdrift.dev/session/F12173F4-D646-47CD-B6DC-2E113784C7F3)

## sample uses

```sh
# omitting some fields set as env vars, like $API_KEY, $APP_ID
cargo run --bin logger-cli -- log \
    --log-level warn --log-type normal \
    --field "key" "value" \
    --field "other key" "other value" \
    "This is a test log"
```

```sh
cargo run --bin logger-cli -- enqueue-artifacts ./*.cap
cargo run --bin logger-cli -- upload-artifacts
```

## todo

* [x] wire up static metadata for app id, platform
* [x] figure out how much delay is actually needed to cleanly start/stop logger (right now its 2s to start and 1s to stop)
* [x] ~~figure out why the network thread is crashing(?)~~ hang on to my shutdown trigger

### DIFF
--- a/logger-cli/src/logger.rs
+++ b/logger-cli/src/logger.rs
@@ -74,17 +74,17 @@ impl LoggerHolder {
     self.logger.shutdown(true);
   }
 
-  pub fn log(&self, cmd: LogCommand, capture_session: bool) {
+  pub fn log(&self, cmd: &LogCommand, capture_session: bool) {
     let session_capture = if capture_session {
       CaptureSession::capture_with_id("cli command")
     } else {
       CaptureSession::default()
     };
     self.logger.new_logger_handle().log(
-      cmd.log_level.into(),
-      cmd.log_type.into(),
-      cmd.message.into(),
-      FieldPairs(cmd.field).into(),
+      cmd.log_level.clone().into(),
+      cmd.log_type.clone().into(),
+      cmd.message.clone().into(),
+      FieldPairs(cmd.field.clone()).into(),
       [].into(),
       None,
       bd_logger::Block::Yes,
--- a/logger-cli/src/main.rs
+++ b/logger-cli/src/main.rs
@@ -5,8 +5,8 @@
 // LICENSE file or at:
 // https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
 
-use crate::cli::Command;
-use crate::logger::{MaybeStaticSessionGenerator, SESSION_FILE};
+use crate::cli::{Command, Options};
+use crate::logger::{LoggerHolder, MaybeStaticSessionGenerator, SESSION_FILE};
 use bd_log::SwapLogger;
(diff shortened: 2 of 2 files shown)


logger-cli/src/logger.rs
```diff
@@ -0,0 +1,156 @@
...
+use std::path::{Path, PathBuf};
+use std::pin::Pin;
+use std::sync::Arc;
+
+pub type LoggerFuture =
+  Pin<Box<dyn Future<Output = anyhow::Result<()>> + 'static + std::marker::Send>>;
+
+pub const SESSION_FILE: &str = "session_id";
+
+#[allow(dead_code)]
```
snowp, line 25: still need this?

logger-cli/src/logger.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    sleep(2000); // some initialization time for the network
+
+    std::thread::spawn(move || {
+      tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(async {
+          handle_unexpected(future.await, "logger top level run loop");
```
snowp, line 58: not sure if handle_unexpected will do anything if we haven't configured an error reporter so this can probably just unwrap or something?

logger-cli/src/logger.rs
```diff
@@ -0,0 +1,156 @@
...
+      None,
+      bd_logger::Block::Yes,
+      session_capture,
+    );
+  }
+}
+
+fn sleep(milliseconds: u64) {
+  std::thread::sleep(std::time::Duration::from_millis(milliseconds));
+}
```
snowp, line 95: a neat pattern here would be something like

```
fn sleep(duration: std::time::Duration) {
  std::thread::sleep(duration)
}
```

then you can call this with `sleep(200.std_milliseconds())` 

using the time::ext::NumericalStdExtension trait

logger-cli/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+    Command::NewSession => {
+      let session_config = sdk_directory.join(SESSION_FILE);
+      std::fs::remove_file(session_config)?;
+    },
+  }
+
+  Ok(())
+}
+
+fn init_logging() {
```
snowp, line 64: you could use SwapLogger from bd_log and it would initialize a bunch of this for you, it's what we use most other places

snowp (approved):
Seems like a good start to me, just a few minor suggestions

PR #224, opened 2025-08-31
api: implement better rate limiting backoff

mattklein123:
1) Remove shutdown handling in the API. It makes the code convoluted and
   doesn't do anything useful. Instead just move the shutdown wait to
   the top level future.
2) Add retry after handling in the API.
3) Move gRPC Code to the codec crate so it can be used in client code. 
4) Some cleanups in order to remove a bunch of clones during config
   processing.

Fixes BIT-5835


mattklein123:
Needs https://github.com/bitdriftlabs/api/pull/73

PR #225, opened 2025-09-02
Crash-resilient key-value store

kstenerud:
This implementation is essentially a hash map backed by a journaled key-value store. All operations are designed to be as quick as possible, while also providing crash resilience. Every modification to the hash map gets stored as a journal entry that can be replayed in order to rebuild the hash map up until that point.

## KV Journal

The journal is a trait with the following implementations:

* In-memory journal, which stores its data in BONJSON format. It maintains a position field in the data file itself, which points to the last complete journal entry written. This allows us to know which data in the file is valid, and reconstruct the key-value map.
* Memory-mapped journal, which is backed by a file on disk. No matter how the app terminates, the data in the buffer will always be flushed to disk. It uses the in-memory journal to store the data to the buffer.
* Double-buffered journal, which keeps two journals internally, forwarding journal API calls to the "active" journal. Whenever the "active" journal reaches its high water mark, the double-buffered journal initializes the "inactive" journal with the current hash map, and then swaps inactve and active journals.

## KV Store

The key-value store holds a double-buffered memory-mapped journal, as well as a local cache of the current hash map state. This speeds up fetches while still persisting changes to disk.

KV journals maintain a timestamp for when they were initialized, which is used by the double-buffered journal to decide which journal should be "active" when reloaded from disk.

### Performance Characteristics

| Operation        | Time Complexity | Notes                           |
|------------------|-----------------|---------------------------------|
| `get()`          | O(1)            | Reads from in-memory cache      |
| `insert()`       | O(1) amortized  | Journal write + cache update    |
| `remove()`       | O(1) amortized  | Journal write + cache update    |
| `contains_key()` | O(1)            | Cache lookup                    |
| `len()`          | O(1)            | Cache size                      |
| `keys()`         | O(1)            | Returns iterator, no allocation |
| `values()`       | O(1)            | Returns iterator, no allocation |
| `as_hashmap()`   | O(n)            | Clones entire cache             |
| `clear()`        | O(1)            | Efficient journal clearing      |

See the README.md for more detailed information.


snowp (commented):
didn't look at the implementation itself yet, so just some high level comments

bd-resilient-kv/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+  ///
+  /// # Errors
+  /// Returns an error if the clearing operation fails.
+  fn clear(&mut self) -> anyhow::Result<()>;
+
+  /// Get the current state of the journal as a `HashMap`.
+  ///
+  /// # Errors
+  /// Returns an error if the buffer cannot be decoded.
+  fn as_hashmap(&mut self) -> anyhow::Result<HashMap<String, Value>>;
```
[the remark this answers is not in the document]
mattklein123: Nevermind I see that you already maintain a cached map of values. I would just make this available by reference. In the double buffered case you can just make the current primary available by reference.

bd-resilient-kv/src/kv_journal/in_memory.rs
```diff
@@ -0,0 +1,446 @@
...
+    }
+    Ok(ratio)
+  }
+
+  /// Create and write metadata object with given timestamp.
+  /// This data is always written starting at offset 17 in the buffer.
+  /// Returns the new current position (17 + metadata length).
+  fn write_metadata(buffer: &mut [u8], timestamp: u64) -> anyhow::Result<usize> {
+    let buffer_len = buffer.len();
+    let mut cursor = &mut buffer[17 ..];
```
mattklein123, line 214: The constructor only validates the buffer is 16 bytes I think? Can't this fail with a bad size? Should the constructor pedantically validate it's >= 17? Might be similar issues elsewhere.

bd-resilient-kv/src/kv_journal/in_memory.rs
```diff
@@ -0,0 +1,446 @@
...
+  }
+
+  fn write_journal_entry(&mut self, key: &str, value: &Value) -> anyhow::Result<()> {
+    let initial_position = self.position;
+    let buffer_len = self.buffer.len();
+    let mut cursor = &mut self.buffer[self.position ..];
+
+    // Write the journal entry using BufMut - position is tracked automatically
+    serialize_map_begin(&mut cursor)
+      .map_err(|e| anyhow::anyhow!("Failed to serialize map begin: {e:?}"))?;
```
mattklein123, line 292: nit: If some of these practically cannot fail I would use the InvariantError to be more clear and also save binary space as these error strings are going to wind up in the binary. Same elsewhere.

bd-resilient-kv/src/kv_journal/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+  ///
+  /// # Errors
+  /// Returns an error if the clearing operation fails.
+  fn clear(&mut self) -> anyhow::Result<()>;
+
+  /// Get the current state of the journal as a `HashMap`.
+  ///
+  /// # Errors
+  /// Returns an error if the buffer cannot be decoded.
+  fn as_hashmap(&mut self) -> anyhow::Result<HashMap<String, Value>>;
```
mattklein123, line 59:
```suggestion
  fn as_hashmap(&self) -> anyhow::Result<&HashMap<String, Value>>;
```

In general it would be more idiomatic to return a reference and then let the caller decide to clone or do whatever. I think this should be pretty easily possible with the way you have it setup? Also I don't think this needs to be mut?

bd-resilient-kv/src/kv_store.rs
```diff
@@ -0,0 +1,246 @@
...
+    self.cached_map.keys()
+  }
+
+  /// Get all values in the store.
+  ///
+  /// This operation returns an iterator over the values without allocation.
+  #[must_use]
+  pub fn values(&self) -> std::collections::hash_map::Values<'_, String, Value> {
+    self.cached_map.values()
+  }
```
mattklein123, line 202: Is any of this used? I would just remove all of this extra stuff if you provide ref access to the hash map, as anything can be done from that.

mattklein123 (changes requested):

bd-resilient-kv/src/kv_journal/in_memory.rs
```diff
@@ -36,6 +34,191 @@ pub struct InMemoryKVJournal<'a> {
...
+
+  // SAFETY: We're writing to the garbage area beyond the valid data position.
+  // This is safe because:
+  // 1. The position is validated to be within buffer bounds by read_position()
+  // 2. We're only writing to buffer[position..] which is the garbage area
+  // 3. The write is temporary and doesn't affect the journal's valid data
+  let buffer_mut =
+    unsafe { std::slice::from_raw_parts_mut(buffer.as_ptr().cast_mut(), buffer.len()) };
+  let mut cursor = &mut buffer_mut[position ..];
+  serialize_container_end(&mut cursor).map_err(|_| InvariantError::Invariant)?;
```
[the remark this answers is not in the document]
kstenerud: Right, I'd forgotten about the partial results. Updated to use that now.

PR #227, opened 2025-09-04
create ANR parsing module

kattrali:
adds a new parser module which uses nom to create a flatbuffers report object from an ANR tombstone. The parser module itself is confined to `lib.rs` and modules in `input.rs` and `android.rs` (with a few tests) while the vast majority of the changeset are fixture files.

- the `input` module defines a wrapper type for memory mapped files so that the parser can use it as a source without pulling the entire tombstone into memory as a string
- `android` defines platform-specific parsers required for the public ANR builder function.
- the test fixture (and snapshot) files are large and extensive to cover several types of complete ANR reports and the many types of information they can contain, like different combinations of frame types, build IDs, and report or thread metadata.

part of BIT-5990

### DIFF
--- a/bd-report-parsers/src/android.rs
+++ b/bd-report-parsers/src/android.rs
@@ -257,11 +257,11 @@ fn build_java_frame<'a, 'fbb, E: ParseError<MemmapView<'a>>>(
       let source_file = Some(v_1::SourceFile::create(builder, &source_file_args));
       let state = state
         .iter()
-        .map(|item| builder.create_string(&item.to_string()))
-        .collect::<Vec<_>>();
+        .map(|item| builder.create_string(item.as_str()))
+        .collect_vec();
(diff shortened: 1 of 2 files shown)


bd-report-parsers/src/android.rs
```diff
@@ -0,0 +1,536 @@
...
+  }
+}
+
+const DEFAULT_ANR_NAME: &str = "Undetermined ANR";
+
+fn anr_name(description: Option<&str>) -> &'static str {
+  let Some(description) = description else {
+    return DEFAULT_ANR_NAME;
+  };
+  let anr_types = vec![
```
snowp, line 502: this could probably be const and just use &[&'static str] everywhere?

snowp (approved):
Not going to pretend like I reviewed the actual parsing logic in detail

PR #228, opened 2025-09-05
improved runtime/workflow crash loop detection

mattklein123:
Now we store the nonce of the last accepted/cached configuration. If we think we have entered a crash loop, we refuse to receive new configuration unless the nonce has changed. In order to support this we no longer clean the cached files when we hit the retry count. The downside of this is that if a client is put into the killed state it would likely come back online and refuse to accept new config if the config hasn't changed. To handle this case we clear cached config when the client enters the killed state.

Fixes BIT-6168

### DIFF
--- a/bd-client-common/src/safe_file_cache.rs
+++ b/bd-client-common/src/safe_file_cache.rs
@@ -176,11 +176,12 @@ impl<T: Message> SafeFileCache<T> {
     // runtime config to accidentally set this really high, but right now this is not
     // configurable at all.
     if retry_count >= MAX_RETRY_COUNT {
-      // Note that eventually if the client is killed this is going to kick in since we attempt
-      // to load cached runtime very early on. Since we don't clean state so that we can do changed
-      // nonce detection below, this could pause killed clients to not get new config when they
-      // come back online since there is no crash loop and the nonce hasn't changed. We handle
-      // this directly in the API code by having the state reset when it enters killed mode.
+      // Note that eventually if the client is killed this is going to kick in since we attempt to
+      // load cached runtime very early on. Since we don't clean state so that we can do changed
+      // nonce detection in the cache_update() function, this could cause killed clients to not get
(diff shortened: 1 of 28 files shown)


bd-client-common/src/safe_file_cache.rs
```diff
@@ -125,28 +152,48 @@ impl<T: Message> SafeFileCache<T> {
...
     // configurable at all.
-    if retry_count > MAX_RETRY_COUNT {
-      // Note that eventually if the client is killed this is going to kick in and delete cached
-      // config. This is OK since we are still covered by the kill file, and ultimately we would
-      // like the client to come up and get fresh config anyway.
-      return Ok((true, None));
+    if retry_count >= MAX_RETRY_COUNT {
+      // Note that eventually if the client is killed this is going to kick in since we attempt
+      // to load cached runtime very early on. Since we don't clean state so that we can do changed
+      // nonce detection below, this could pause killed clients to not get new config when they
```
snowp, line 181: What does "below" refer to here? We're just returning in this function

bd-client-common/src/safe_file_cache_test.rs
```diff
@@ -0,0 +1,77 @@
...
+      ClientKillFile::default(),
+      cache.handle_cached_config().await.unwrap()
+    );
+  }
+
+  // The next time we come up we should refuse to load the cached config.
+  let cache = SafeFileCache::<ClientKillFile>::new("test", tempdir.path());
+  assert!(cache.handle_cached_config().await.is_none());
+
+  // We should need delete any state so make sure we come up again but get no config.
```
snowp, line 49:
```suggestion
  // We should delete any state so make sure we come up again but get no config.
```

snowp (approved):

bd-client-common/src/safe_file_cache.rs
mattklein123: Will clarify

PR #231, opened 2025-09-09
part 1 of workflow debugger

mattklein123:
This does not implement "ackless" streaming log uploads. I will do that in a follow up independent change.


mattklein123:
Depends on https://github.com/bitdriftlabs/api/pull/74. Leaving in draft as there are likely going to be further API changes as I sort out what we need on the server.

bd-workflows/src/engine.rs
```diff
@@ -319,11 +351,12 @@ impl WorkflowsEngine {
...
       let workflow = existing_workflow_ids_to_workflows
-        .remove(config.id())
+        .remove(config.inner().id())
         .map_or_else(
+          // No cached state for a given workflow ID, start from scratch.
+          || Workflow::new(config.inner().id().to_string()),
           // We have cache state for a given workflow ID.
-          || Workflow::new(config.id().to_string()),
-          //   No cached state for a given workflow ID, start from scratch.
+          // fixfix cleanup old debug state?
```
mattklein123, line 359: implement/test this.

bd-workflows/src/engine.rs
```diff
@@ -586,9 +628,17 @@ impl WorkflowsEngine {
         self.needs_state_persistence = true;
       }
 
-      let (triggered_actions, workflow_logs_to_inject) = result.into_parts();
-      actions.extend(triggered_actions);
+      // In debug only mode we do not trigger any actions, but we still inject logs so that
+      // workflows continue to advance if they depend on the injected logs.
```
snowp, line 632: Agreed that this is required for correctness, though interesting how we do end up having debug workflows have a real impact on the timeline data

snowp (approved):

PR #245, opened 2025-09-15
logger streaming mode changes

mattklein123:
1) Use ackless uploads to increase throughput at the expense of
   losing data. Logs still go through rate limiting and back
   pressure will be applied on the API so client memory is
   bounded.
2) Making the streaming uploader use BatchBuilder and have a
   flush deadline which defaults to 500ms. This should in general
   increase payload sizes and keep things snappy.

Fixes BIT-6338

### DIFF
--- a/logger-cli/Cargo.toml
+++ b/logger-cli/Cargo.toml
@@ -2,36 +2,31 @@
 edition      = "2024"
 license-file = "../LICENSE"
 name         = "logger-cli"
-version      = "1.0.0"
 publish      = false
+version      = "1.0.0"
 
 [dependencies]
-anyhow.workspace = true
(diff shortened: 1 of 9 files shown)


logger-cli/Cargo.toml
```diff
@@ -2,36 +2,31 @@
...
+tokio.workspace              = true
 tracing-subscriber.workspace = true
 
 [dev-dependencies]
 tempfile.workspace = true
-
-# improved rust-lldb accessibility:
-[profile.dev]
-debug = true
-opt-level = 0
```
mattklein123, line 37: Per warning I think this is ignored in workspace mode.

snowp (approved):

PR #246, opened 2025-09-16
API shell for feature flags

kstenerud:


bd-logger/src/builder.rs
```diff
@@ -179,6 +180,34 @@ impl LoggerBuilder {
     let (crash_monitor_tx, crash_monitor_rx) = tokio::sync::oneshot::channel();
 
     let network_quality_provider = Arc::new(SimpleNetworkQualityProvider::default());
+
+    let feature_flags_manager = FeatureFlagsManager::new(
```
snowp, line 184: Would it be hard to move all this into the `async {}` block and pass the write side of the FF to async_log_buffer in its start() call? I think that would let us do all these ops on the event loop and avoid inline IO and also make it easier to hook this all up to the runtime flags that can let us dynamically configure the parameters etc

bd-logger/src/async_log_buffer.rs
```diff
@@ -353,6 +359,17 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
     ))
   }
 
+  pub fn set_feature_flag(
+    tx: &Sender<AsyncLogBufferMessage>,
+    flag: &str,
```
snowp, line 364: since we immediately call to_string() on the flag it can be nice to have the function take String so the caller can decide whether to move or copy, the current interface forces a copy

bd-logger/src/logger.rs
```diff
@@ -470,6 +488,9 @@ pub struct InitParams {
   // Whether the logger should start in sleep mode. It can then be transitioned using the provided
   // transition APIs.
   pub start_in_sleep_mode: bool,
+
+  pub feature_flags_file_size_bytes: usize,
+  pub feature_flags_high_watermark: f32,
```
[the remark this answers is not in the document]
kstenerud: Yes, the idea for now is to get this running, and then sort out how to do these bits better.

kstenerud:
We now have a "remove flag" API to match parity with other similar libraries.

bd-logger/src/builder.rs
```diff
@@ -179,6 +180,15 @@ impl LoggerBuilder {
...
 
     let network_quality_provider = Arc::new(SimpleNetworkQualityProvider::default());
+
+    let feature_flags_builder = FeatureFlagsBuilder::new(
+      &self.params.sdk_directory,
+      self.params.feature_flags_file_size_bytes,
+      self.params.feature_flags_high_watermark,
+    );
+    // This is only a delete and rename, so it's fast
+    feature_flags_builder.backup_previous();
```
snowp, line 190: Since both the reader and the writer of this won't interact with this until they are started on the event loop I think we could just move this line inside
```
    let logger_future = async move {
```
and have it happen on the bg thread?

bd-logger/src/async_log_buffer.rs
```diff
@@ -661,6 +684,15 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
...
     // the async log buffer goes back to processing incoming logs.
 
+    let mut current_feature_flags = self
+      .feature_flags_builder
+      .current_feature_flags()
+      .inspect_err(|e| {
+        // This should never fail unless there's a serious filesystem issue.
+        log::warn!("failed to load or create current feature flags: {e}");
+      })
+      .ok();
```
snowp, line 694: Should we have this be lazily constructed in response to the first feature flag call we get? That way this is not initialized unless they make use of feature flags

Since this does file I/O we may want to wrap it in a `tokio::task::spawn_blocking` so that the I/O doesn't interfere with the event loop

snowp (commented):
Nice this is looking pretty close, just a few things

snowp (approved):
Nice!

PR #251, opened 2025-09-17
various fixed to logger-cli

snowp:
- Handle the case when there is no session ID file
- Fix calling process_crash_report from a tokio runtime. Using spawn_blocking we can reuse the blocking logic
- Default server addr to locahost which seems to work better with ipv6


logger-cli/src/logger.rs
```diff
@@ -179,6 +183,7 @@ pub fn make_logger(sdk_directory: &Path, config: &StartCommand) -> anyhow::Resul
     ))),
     metadata_provider: Arc::new(LogMetadata {
       timestamp: time::OffsetDateTime::now_utc().into(),
+      ootb_fields: [("_app_version_code".into(), "10".into())].into(),
```
snowp, line 186: this needs to be set in order for lapi to accept the artifact upload, maybe this should be configurable via a cli arg instead?

PR #254, opened 2025-09-22
fix(bd-logger): avoid interpolating current and prev session metadata

kattrali:
Removes normalization pass for previous session logs, which may have had an entirely different global state. Avoids issues particularly in crash handling where fields like app version differs from the current version of the app and pre-populated data from cached previous session global session and the crash report is overwritten by current-moment data from metadata providers

### DIFF
--- a/bd-logger/src/async_log_buffer.rs
+++ b/bd-logger/src/async_log_buffer.rs
@@ -115,6 +115,9 @@ pub enum LogAttributesOverrides {
   /// The hint that tells the SDK what the expected previous session ID was. The SDK uses it to
   /// verify whether the passed information matches its internal session ID tracking and drops
   /// logs whose hints are invalid.
+  ///
+  /// Use of this override assumes that all relevant metadata has been attached to the log as no
+  /// current session metadata will be added.
   PreviousRunSessionID(String, OffsetDateTime),
 
   /// Overrides the time when the log occurred at, useful for cases like spans with a provided


bd-logger/src/async_log_buffer.rs
```diff
@@ -447,14 +447,24 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
...
-          log.fields,
-          log.matching_fields,
-          log.log_type,
-          &mut self.global_state_tracker,
-        )
+      if let Some(LogAttributesOverrides::PreviousRunSessionID(_id, _timestamp)) =
+        &log.attributes_overrides
+      {
+        // avoid normalizing metadata for logs from previous sessions, which may
+        // have had different global state
```
snowp, line 454: I think this assumes that `PreviousRunSessionId` means that the caller handles global state but not sure if that's always true? This API is used by the AppExitLogger which I don't think handles the old global state, and we also probably need to think about whether we want the crash logs to be "naked" (aka no basic fields like app_version etc.) in the case where we failed to recover the old global state

bd-logger/src/async_log_buffer.rs
```diff
@@ -447,14 +447,24 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
...
     let result = with_thread_local_logger_guard(|| {
-      self
-        .metadata_collector
-        .normalized_metadata_with_extra_fields(
-          log.fields,
-          log.matching_fields,
-          log.log_type,
-          &mut self.global_state_tracker,
-        )
+      if let Some(LogAttributesOverrides::PreviousRunSessionID(_id, _timestamp)) =
```
snowp, line 450: Maybe drop a note on the PreviousRunSessionID enum about this behavior? The interface doesn't make it obvious that the way fields are handled changes so might be good to document it at least

snowp (approved):
Nice, make sense with the capture-sdk change

PR #255, opened 2025-09-22
bd-key-value's Storage trait, applied to KVStore

kstenerud:
This PR adds a new `Storage` that uses `bd_resilient_kv` under the hood, which gives us crash-resilient storage of the globals.

This allows us to move off the `Preferences` (Android) and `UserDefaults` (iOS) implementations on mobile.

### DIFF
--- a/bd-key-value/src/resilient_kv.rs
+++ b/bd-key-value/src/resilient_kv.rs
@@ -1,0 +1,105 @@
+// shared-core - bitdrift's common client/server libraries
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+#![deny(
+  clippy::expect_used,
+  clippy::panic,
+  clippy::todo,
+  clippy::unimplemented,
+  clippy::unreachable,
+  clippy::unwrap_used
+)]
+
+use crate::Storage;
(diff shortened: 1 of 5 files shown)


bd-key-value/src/resilient_kv.rs
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  fn with_store<F, R>(&self, f: F) -> anyhow::Result<R>
+  where
+    F: FnOnce(&KVStore) -> anyhow::Result<R>,
+  {
+    let store = self
+      .store
+      .lock()
+      .map_err(|e| anyhow::anyhow!("Failed to acquire lock: {}", e))?;
```
mattklein123, line 70: use parking_lot::Mutex to avoid all this error checking.

mattklein123 (changes requested):
LGTM with small nit.

bd-key-value/src/resilient_kv.rs
```diff
@@ -0,0 +1,105 @@
...
+  fn with_store_mut<F, R>(&self, f: F) -> anyhow::Result<R>
+  where
+    F: FnOnce(&mut KVStore) -> anyhow::Result<R>,
+  {
+    let mut store = self.store.lock();
+    f(&mut store)
+  }
+}
+
+impl Storage for ResilientKvStorage {
```
snowp, line 81: We can do this later but now that there is a single implementation for Storage we should probably improve the Storage API a bit so that we can store more than just string types. For example the global fields are stored as a map serialized into a string but seems like we should be able to use better types within the storage at this point?

PR #270, opened 2025-10-06
Support adding multiple feature flags at once

kstenerud:
I've also removed the callback mechanism because it was never used, and it would complicate things when calling set multiple.

### DIFF
--- a/bd-feature-flags/src/lib.rs
+++ b/bd-feature-flags/src/lib.rs
@@ -257,7 +257,8 @@ impl FeatureFlags {
   pub fn set_multiple(&mut self, flags: Vec<(String, Option<String>)>) -> anyhow::Result<()> {
     // Convert the input vector to Vec format for the KV store
     let now = time::OffsetDateTime::now_utc();
-    let kv_entries: Vec<(String, bd_bonjson::Value)> = flags.into_iter()
+    let kv_entries: Vec<(String, bd_bonjson::Value)> = flags
+      .into_iter()
       .map(|(key, variant)| {
         let feature_flag = FeatureFlag::new(variant, Some(now))?;
         let value = feature_flag.to_value();
--- a/bd-feature-flags/src/lib.rs
+++ b/bd-feature-flags/src/lib.rs
@@ -230,6 +230,45 @@ impl FeatureFlags {
     Ok(())
   }
 
+  /// Sets or updates multiple feature flags in a single operation.
+  ///
+  /// Creates or updates multiple feature flags with their respective variants. All flags are
+  /// immediately stored in persistent storage and receive timestamps indicating when they were
+  /// last modified. This method is more efficient than calling `set()` multiple times.
+  ///
+  /// # Arguments
+  ///
+  /// * `flags` - A map of flag names to their variants:
+  ///   - `Some(string)` sets the flag with the specified variant
+  ///   - `Some("")` treats the variant as `None`
+  ///   - `None` sets the flag without a variant (simple boolean-style flag)
+  ///
+  /// # Returns
+  ///
+  /// Returns `Ok(())` on success, or an error if any flag cannot be stored.
+  ///
+  /// # Errors
+  ///
+  /// This function will return an error if:
+  /// - Any flag cannot be written to persistent storage due to I/O errors, insufficient disk space,
+  ///   or permission issues
+  /// - If an error occurs, some flags may have been successfully written while others may not have
+  ///   been processed
+  pub fn set_multiple(&mut self, flags: &HashMap<String, Option<&str>>) -> anyhow::Result<()> {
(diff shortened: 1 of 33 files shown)


bd-feature-flags/src/lib.rs
```diff
@@ -230,6 +229,44 @@ impl FeatureFlags {
...
+  ///
+  /// Returns `Ok(())` on success, or an error if any flag cannot be stored.
+  ///
+  /// # Errors
+  ///
+  /// This function will return an error if:
+  /// - Any flag cannot be written to persistent storage due to I/O errors, insufficient disk space,
+  ///   or permission issues
+  /// - If an error occurs, no flags will be written.
+  pub fn set_multiple(&mut self, flags: &HashMap<String, Option<&str>>) -> anyhow::Result<()> {
```
mattklein123, line 255: I don't think there is any compelling reason to have this take a hashmap, just to convert to another hashmap and then pass into the store. I would have this take a slice of tuples, which you can then convert into a Vec if needed, and then pass into the insert_multiple function. If someone passes duplicates it would end up getting resolved on final insert.

bd-resilient-kv/src/kv_journal/in_memory.rs
```diff
@@ -324,6 +317,31 @@ impl<'a> InMemoryKVJournal<'a> {
     self.set_position(buffer_len - remaining);
     Ok(())
   }
+
+  fn write_journal_entries(&mut self, entries: &HashMap<String, Value>) -> anyhow::Result<()> {
+    let buffer_len = self.buffer.len();
+    let mut cursor = &mut self.buffer[self.position ..];
+
+    // Try to write all entries as a single BONJSON object
+    let encode_result = encode_into_buf(&mut cursor, &Value::Object(entries.clone()));
```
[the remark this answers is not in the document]
mattklein123: +1 there is no reason to clone here.

mattklein123 (changes requested):

bd-feature-flags/src/lib.rs
kstenerud: At the bottom level, it needs to be written as a map because that's how the journal stores a journal entry: As a map containing keys and values that need to be added/replaced on replay.

No matter what it will need to somewhere along the line convert whatever we've passed in to a map type that's suitable for serialization (`HashMap<String, Value>`).

Alternatively we could fake it out by writing a manual `Object` start and end marker, and then walk the slice, adding in key-value entries as we go. But then we'd have to duplicate all of the serialization code that handles containers in depth.

kstenerud: Actually, we could make the `FeatureFlags` API take a slice of `(String, FeatureFlag)`, and then have it converted to a hash map in this function. We'd still need to decide what the native side API should look like...

bd-feature-flags/src/lib.rs
```diff
@@ -252,15 +253,15 @@ impl FeatureFlags {
   /// - Any flag cannot be written to persistent storage due to I/O errors, insufficient disk space,
   ///   or permission issues
   /// - If an error occurs, no flags will be written.
-  pub fn set_multiple(&mut self, flags: &HashMap<String, Option<&str>>) -> anyhow::Result<()> {
-    // Convert the input HashMap to the format expected by the KV store
+  pub fn set_multiple(&mut self, flags: Vec<(String, Option<String>)>) -> anyhow::Result<()> {
+    // Convert the input vector to the format expected by the KV store
     let mut kv_entries = HashMap::new();
```
mattklein123, line 258: You don't need a hashmap here either. All you do is take this and then iterate over it internally in the store. The store should take a slice/vec as well.

mattklein123 (changes requested):

bd-feature-flags/src/lib.rs
kstenerud: `self.flags_store` is a `KVStore`, which doesn't understand the concept of feature flags, so it can't set up the `key:(variant, timestamp)` structure.

I could convert it here to a vec of `(String, Object(HashMap<String, Value>))`, but it seems wasteful to make a vec, only to immediately convert it into the hash map that's here already now (the final form it must take in order to be serialized).

mattklein123: Inside the store you are iterating through and just inserting items. Meaning, you are converting to a hashmap just to convert again. This is wasteful. Am I missing something. If you need to convert the Vec that is still much faster.

mattklein123: Per offline convo let's just add a TODO and use `AHashMap` throughout all this code for now.

kstenerud: I've updated it now to pass vectors all along the line.

mattklein123 (approved):
Nice thanks,

PR #304, opened 2025-11-03
feature flags: emit logs on each feature flag operation 

snowp:
This emits a log every time one of the feature flag operations are performed, allowing for workflows to match on these events

Currently the log format is a bit of a strawman proposal, subject to change as we discuss

### DIFF
--- a/bd-logger/src/async_log_buffer/feature_flags.rs
+++ b/bd-logger/src/async_log_buffer/feature_flags.rs
@@ -47,7 +47,7 @@ impl FeatureFlagsHolder {
   pub async fn set(&mut self, flag: String, variant: Option<&str>) -> LogLine {
     if let Some(feature_flags) = self.maybe_initialize_feature_flags().await {
       feature_flags
-        .set(flag.clone(), variant.as_deref())
+        .set(flag.clone(), variant)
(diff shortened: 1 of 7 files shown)


bd-logger/src/async_log_buffer/feature_flags.rs
```diff
@@ -0,0 +1,182 @@
...
+      flags
+        .into_iter()
+        .map(|(flag, variant)| {
+          (
+            format!("_set_flag_{flag}").into(),
+            AnnotatedLogField::new_ootb(variant.unwrap_or_else(|| "none".to_string())),
+          )
+        })
+        .collect(),
+    )
```
mattklein123: If we add a new log type, IMO we should use a binary format for efficiency (similar to session replay). Also, not sure if we should consider chunking super large batch operations? Like what if we have the 2K set case, should we send 2K in the single log line? It could get pretty big.

The downside of course of the binary format is that it makes matching much more difficult. I wonder if we should keep some dedicated map for matching, and then before it's serialized the buffer convert this to binary format? the on demand serialization would also be required for field compression so possibly some shared machinery there?

mattklein123 (commented):
Some high level comments for discussion.

PR #309, opened 2025-11-06
store crash_monitor instance from transaction

kattrali:
Resolves BIT-6588

preserves the oneshot transaction within the logger builder for sending the monitor instance, which allows the logger to fetch it and store it for multiple uses. this should keep the logger internals private (and nothing else has to know how it handles the monitor internally)

also rolls back some unrelated changes to make the intent clearer

### DIFF
--- a/bd-logger/src/logger.rs
+++ b/bd-logger/src/logger.rs
@@ -38,7 +38,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use time::ext::NumericalDuration;
 use tokio::sync::mpsc::{Receiver, Sender};
-use tokio::sync::watch;
+use tokio::sync::{oneshot, watch};
 
 #[derive(Clone)]
 #[allow(clippy::struct_field_names)]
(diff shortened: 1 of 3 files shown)


bd-logger/src/logger.rs
```diff
@@ -676,16 +681,24 @@ impl Logger {
...
-      }
-      std::thread::sleep(Duration::from_millis(10));
+    let crash_monitor = if let Some(monitor) = self.crash_monitor.lock().clone() {
+      monitor
+    } else if let Some(rx) = self.crash_monitor_rx.take() {
+      let monitor = rx.blocking_recv()?;
+      *self.crash_monitor.lock() = Some(monitor.clone());
+      monitor
+    } else {
+      anyhow::bail!("failed to store/receive crash_monitor instance");
```
kattrali: This is the sauce of the pr: used the cached monitor if there is one, otherwise get it from the builder transaction and store it, or else bail as something has gone wrong

FranAguilera (approved):
Works great! Verified locally with this session that triggers non-fatals and uses `ReportProcessingSession:Current` to report without app restart. See verified [session here](https://timeline.bitdrift.dev/session/b0c0abf0-cdf2-4362-9bab-5f18b2854d25?utm_source=sdk&utilization=0) 

- [Non-fatals here ](https://timeline.bitdrift.dev/session/b0c0abf0-cdf2-4362-9bab-5f18b2854d25?utm_source=sdk&utilization=0&expanded=6353188582093526929)(btw, ignore the "Fatal text")

- [Fatal ANR here](https://timeline.bitdrift.dev/session/b0c0abf0-cdf2-4362-9bab-5f18b2854d25?utm_source=sdk&utilization=0&expanded=-5838603472493663890)

PR #445, opened 2026-04-15
[Battery Level Change] Allow window configuration

FranAguilera:
## What

Part of BIT-7976

- Updating the `Battery Level Change` window to be configurable (now it's hardcoded to 1 minute)
- Keep window default as 1 minute

## Verification

Verified with emulator session

## Follow-up

Frontend will be updated to handle the new field name

### DIFF
--- a/bd-logger/src/battery.rs
+++ b/bd-logger/src/battery.rs
@@ -19,23 +19,37 @@ use bd_log_primitives::{
   LogMessage,
 };
 use bd_proto::protos::logging::payload::LogType;
+use bd_runtime::runtime::resource_utilization::BatteryLevelChangeWindowFlag;
+use bd_runtime::runtime::{ConfigLoader, Watch};
 use std::collections::VecDeque;
 use std::sync::Arc;
-use time::ext::NumericalDuration;
 
 //
 // BatteryDrainTracker
 //
 
 /// Tracks battery level changes over time by monitoring battery level changes in RESOURCE logs.
-/// Calculates the percentage point change per minute and adds it to resource logs.
+/// Calculates the percentage point change over a configurable window and adds it to resource logs.
 pub struct BatteryDrainTracker {
+  battery_level_change_window: Watch<time::Duration, BatteryLevelChangeWindowFlag>,
   container: parking_lot::Mutex<BatteryMetricsContainer>,
 }
 
 impl BatteryDrainTracker {
-  pub fn new(time_provider: Arc<dyn TimeProvider>) -> Self {
+  pub fn new(time_provider: Arc<dyn TimeProvider>, runtime: &ConfigLoader) -> Self {
     Self {
+      battery_level_change_window: runtime.register_duration_watch(),
+      container: parking_lot::Mutex::new(BatteryMetricsContainer::new(time_provider)),
+    }
+  }
+
+  #[cfg(test)]
+  fn new_with_window(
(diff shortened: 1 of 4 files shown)


bd-runtime/src/runtime.rs
```diff
@@ -795,6 +795,12 @@ pub mod resource_utilization {
     "resource_utilization.reporting_interval_ms",
     6.seconds()
   );
+
+  duration_feature_flag!(
+    BatteryLevelChangeWindowFlag,
+    "resource_utilization.battery_level_change_window_seconds",
```
murki: it says seconds but it should be minutes

bd-runtime/src/runtime.rs
```diff
@@ -795,6 +795,12 @@ pub mod resource_utilization {
...
+    5.minutes()
```
murki: talked offline and we agreed that the default should probably still be 1

murki:
I think it makes sense to make it runtime configurable with some sensible default. I'll defer to @snowp for the actual code review

bd-logger/src/battery.rs
```diff
@@ -44,17 +60,18 @@ impl BatteryDrainTracker {
...
     }
 
-    if let Some(drain_rate) = guard.get_level_change_per_minute() {
+    if let Some(battery_level_change) = guard.get_level_change(battery_level_change_window) {
       fields.insert(
-        "_battery_level_change_per_min".into(),
-        AnnotatedLogField::new_ootb(format!("{drain_rate:.4}")),
+        "_battery_level_change".into(),
+        AnnotatedLogField::new_ootb(format!("{battery_level_change:.4}")),
       );
```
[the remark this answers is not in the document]
snowp: this seems legit, did you want to change the key?

bd-logger/src/battery.rs
```diff
@@ -127,6 +144,22 @@ impl BatteryMetricsContainer {
   }
 }
 
+enum BatteryLevelChangeWindow {
+  Runtime(Watch<time::Duration, BatteryLevelChangeWindowFlag>),
+  #[cfg(test)]
+  Fixed(time::Duration),
```
snowp: Instead of this I think you can just do `Watch::new_for_testing` to pass in a fixed ff flag

bd-runtime/src/runtime.rs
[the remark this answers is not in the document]
snowp: ^legit, if you want to use duration_feature_flag! it must be given in ms. We should add a different one that would allow other units but simplest for now is to have this take ms

bd-runtime/src/runtime.rs
FranAguilera: Updated

bd-logger/src/battery.rs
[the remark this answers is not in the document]
FranAguilera: Yes, that was intentional given the key will lost its meaning if we change the window to other value (e.g. 5 minutes). I was thinking to update later loop-api/frontend to handle this on newer versions

bd-runtime/src/runtime.rs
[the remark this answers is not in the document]
FranAguilera: good point, will update to ms

bd-logger/src/battery.rs
FranAguilera: will do

murki (approved):

PR #476, opened 2026-04-28
Couple stats scheduling and add app-open testing hooks

mattklein123:
## Summary
- replace the client stats dual-ticker flow with a shared periodic schedule that flushes before uploads
- preserve upload deadlines across flush-only runtime changes, upload old pending data on startup, and narrow the minimum upload gate to flush-triggered uploads
- wire the new schedule through `bd-client-stats`, `bd-logger`, profiling, and runtime config comments
- add dedicated `RuntimePeriodicSchedule` coverage plus startup and in-flight upload behavior tests
- add logger-cli stats observation support and the app-start-testing example script used by the open-testing workflow


bd-client-stats/src/observer.rs
```diff
@@ -0,0 +1,67 @@
...
+
+pub trait StatsObserver: Send + Sync {
+  fn on_snapshot(&self, observation: SnapshotObservation);
+
+  fn on_upload_attempt(&self, observation: UploadAttemptObservation);
+
+  fn on_upload_ack(&self, observation: UploadAckObservation);
+}
+
+static OBSERVER: OnceLock<RwLock<Option<Arc<dyn StatsObserver>>>> = OnceLock::new();
```
snowp: I think LazyLocks saves you the explicit get_or_init command

bd-client-stats/src/stats.rs
```diff
@@ -37,148 +47,333 @@ use itertools::Itertools;
...
+}
+
+// Couples the stats flush and upload timers into one schedule driven by the active upload
+// interval. The scheduler always reasons about one upload cycle at a time and may insert one or
+// more intermediate flush deadlines before the next upload deadline when the configured flush
+// interval is a clean divisor of the active upload interval.
+//
+// Startup is deterministic: the first cycle is scheduled without jitter. We only re-jitter upload
+// deadlines when the active upload interval changes at runtime, which keeps reconnect storms from
+// synchronizing without making every process immediately flush or upload on boot.
```
snowp: So we still wait for the interval on startup? Or do we immediately schedule? The last part makes it sound like we aren't immediately trying to upload

FranAguilera (approved):

logger-cli/examples/app-start-testing.sh
snowp, line 1: nice python script

snowp (approved):

logger-cli/examples/app-start-testing.sh
mattklein123: Yes, yes. I will translate it.

bd-client-stats/src/stats.rs
mattklein123: We do try to flush on startup now I will make this more clear.

bd-client-stats/src/observer.rs
mattklein123: Will fix.

PR #479, opened 2026-05-01
reports: add scripting engine

kattrali:
Adds a new execution engine (`Script`) and trait (`Scriptable`) to support passing arbitrary structures as script input. 

### Compiling

The core component of the engine is implementing the `Scriptable` trait:

```rs
/// Base definition for objects passable to `Script::run()`
pub trait Scriptable {
  /// Return a value for a (potentially) nested value or self when empty
  fn resolve(&self, path: &[OwnedSegment]) -> Result<Option<ScriptValue>, PathError>;

  /// Define type definitions which will be used as hints within the scripting
  /// engine
  fn schema() -> Kind;
}
```

Given a particular `Scriptable` type, a new script can be compiled from source code. The generic type provides type hints to the compiler for what fields are available.

```rs
let script = Script::new::<ScriptableThing>(text, vec![ /* custom functions */ ])?;
```

Type definitions are provided for common base types (numbers, strings, bool, vectors of any of those, `serde_json` objects, etc) + generated implementations of the core `Report` types based on the flatbuffer schema (most of the code tbh).

### Running

A compiled script can then be run against an instance of `ScriptableThing`:

```rs
let thing = ScriptableThing { ... };
let output = script.run::<MyOutputType>(&thing)?;
```

### Functions

Custom functions for the script runtime can be used to define the result of running the script. 

```rs
struct MyOutputType {
  text: String
}

// snipped from a FunctionExpression impl:
if let Some(data) = get_dynamic_data::<MyOutputType>(ctx) else {
  data.text = "something from the script".to_string();
};
```

#### Provided functions

if `report_functions()` is specified to `Script::new()`:

* `add_field(name, value)` - appends a value which will then be included in report insights
* `set_significant_frame(error_index, frame_index, bool)` - mark a particular frame as important (or unimportant) to the overall issue (with the bool value defaulting to `true`). example:
  ```vrl
   for_each(.errors) -> |error_index, error| {
      for_each(error.stacktrace) -> |frame_index, frame| {
        if starts_with(frame, "os.") || contains(frame, ".rx.") {
          set_significant_frame(error_index, frame_index, false)
        } else if starts_with(frame, "com.myapp.") {
          set_significant_frame(error_index, frame_index)
        }
      }
   }
  ```
* `set_grouping_key(value)` - set the sole determinant for grouping a particular issue, ignoring other factors like stack trace contents


bd-script/src/report/functions/add_field.rs
```diff
@@ -0,0 +1,85 @@
...
+  Value,
+  kind,
+};
+
+#[derive(Debug)]
+pub struct AddField;
+
+impl Function for AddField {
+  fn identifier(&self) -> &'static str {
+    "add_field"
```
mattklein123, line 26: We should probably prefix all our custom functions with "bitdrift" or "bdrl" or whatever mostly just to avoid any future conflicts with the standard libary.

bd-script/src/functions/add_field.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl FunctionExpression for FieldData {
+  fn resolve(&self, ctx: &mut vrl::prelude::Context<'_>) -> vrl::prelude::Resolved {
+    let name_value = self.name.resolve(ctx)?;
+    let Some(name) = name_value.as_str() else {
+      return Err(ExpressionError::from("name must be a string value"));
+    };
+    let value = self.value.resolve(ctx)?;
+    ctx
```
mattklein123: You likely already thought about this but obviously to do this we don't really need a function, we could just handle writing to `.fields` or whatever. I think having a function is probably clearer and that's fine with me but just throwing it out there.

bd-script/src/functions/add_field.rs
```diff
@@ -0,0 +1,85 @@
...
+  fn parameters(&self) -> &'static [Parameter] {
+    &[
+      Parameter {
+        keyword: "name",
+        kind: kind::BYTES,
+        required: true,
+      },
+      Parameter {
+        keyword: "value",
+        kind: kind::ANY,
```
mattklein123: Should we just force a string here also for now?

bd-script/src/functions/set_grouping_key.rs
```diff
@@ -0,0 +1,76 @@
...
+  fn resolve(&self, ctx: &mut vrl::prelude::Context<'_>) -> vrl::prelude::Resolved {
+    let key = self.key.resolve(ctx)?;
+    if !key.is_bytes() && !key.is_null() {
+      return Err(ExpressionError::from(
+        "grouping key must be a string value or null",
+      ));
+    }
+    ctx
+      .target_mut()
+      .target_insert(&grouping_key_path(), key.clone())?;
```
mattklein123: You may have saw this already, but I built a way where the functions can store data out of band from the main Value construct. It doesn't really matter but just letting you know if this ends up being easier.

So basically you can access a real object without having to go through this indirection: https://github.com/bitdriftlabs/pulse/blob/main/pulse-metrics/src/vrl/pulse_add_to_downstream_id.rs#L84-L92

bd-script/src/report/generated/mod.rs
```diff
@@ -0,0 +1,1912 @@
+// shared-core - bitdrift's common client/server libraries
```
mattklein123, line 1: Is this generated? Where does this come from? I don't see the build.rs that I think you showed me.

bd-script/src/input.rs
```diff
@@ -0,0 +1,501 @@
...
+      ))
+    }
+  }
+
+  fn schema() -> Kind {
+    Kind::boolean()
+  }
+}
+
+impl Scriptable for f32 {
```
mattklein123: nit: you could probably reduce a lot of boilerplate in here with some macros.

bd-script/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+  pub fn run<T: Scriptable>(
+    &self,
+    object: &T,
+    feature_flags: &[FeatureFlag],
+  ) -> anyhow::Result<Output> {
+    let mut state = RuntimeState::default();
+    let timezone = TimeZone::default();
+    let mut target = ReportTarget::new(object, feature_flags);
+    let mut ctx = Context::new(&mut target, &mut state, &timezone);
+    self.program.resolve(&mut ctx)?;
```
mattklein123, line 69: When you wire this up we are likely going to have to differentiate between abort() and outright failure so you probably need thiserror or some other way to map. When we use this for matching we will want abort() to signify no match and probably handle that different from an outright error from a monitoring perspective.

bd-script/src/target.rs
```diff
@@ -0,0 +1,157 @@
...
+use std::collections::BTreeMap;
+use vrl::compiler::{OwnedValueOrRef, SecretTarget, Target};
+use vrl::core::Value;
+use vrl::path::{OwnedSegment, OwnedTargetPath, PathPrefix};
+
+const GROUPING_KEY: &str = "grouping_key";
+const ERRORS_KEY: &str = "errors";
+const METRICS_KEY: &str = "metrics";
+const STACKTRACE_KEY: &str = "stack_trace";
+const IS_SIGNIFICANT_KEY: &str = "is_significant";
```
mattklein123: Per my other comment it might be cleaner to go completely out of band of the object but up to you.

mattklein123 (commented):
Super cool!

bd-script/src/functions/add_field.rs
kattrali: Yeah - I started there and then made it more structured with functions so we can provide better error messages

bd-script/src/report/functions/abort.rs
```diff
@@ -0,0 +1,83 @@
+// shared-core - bitdrift's common client/server libraries
```
mattklein123: You don't need an abort function. This is built into VRL. The script will return with a special status code that tells you it was aborted.

bd-script/src/report/functions/add_field.rs
mattklein123: ping on this, same elsewhere

mattklein123 (changes requested):
lgtm just some small remaining comments.

bd-script/src/report/functions/add_field.rs
kattrali: actually meant to follow up on this - do we actually need that if its going to be its own language/thing?

mattklein123: It's probably fine either way, it's mostly a matter of whether we want to be able to easily merge in from upstream in terms of both the docs and stdlib functions. Obviously since we are forked we can change the upstream name as well. Will leave it to you to decide.

.github/workflows/rust-library.yaml
```diff
@@ -11,6 +11,7 @@ concurrency:
 env:
   CARGO_NET_GIT_FETCH_WITH_CLI: true
   CARGO_TERM_COLOR: always
+  SKIP_FILE_GEN: true
```
mattklein123, line 14: These changes are going to get reverted, same in the other file. This should go in the Makefile.

mattklein123 (approved):

bd-script/src/report/functions/add_field.rs
kattrali: going with assuming its fine, and will handle prefixing from upstream if needed

PR #481, opened 2026-05-02
Implement new session APIs

mattklein123:
## Summary
Implement the shared-core side of the new session API changes.

- add the new client API/session update support and regenerate the affected proto bindings
- rework `bd-session` persistence, mutation flow, and tests for the async session model
- propagate the new session behavior through `bd-api`, `bd-logger`, `bd-crash-handler`, `bd-error-reporter`, and `logger-cli`
- keep the SDK-facing logger session APIs synchronous by bridging requests through the async logger task

Fixes BIT-8150


mattklein123:
Depends on https://github.com/bitdriftlabs/api/pull/147. 

Most of the diff is proto changes.

bd-logger/src/test/logger_integration.rs
```diff
@@ -524,7 +537,12 @@ fn api_bandwidth_counters() {
       // If these numbers end up being too variable we do something more generic.
       let bandwidth_tx = upload.get_counter("api:bandwidth_tx", labels! {}).unwrap();
       let bandwidth_rx = upload.get_counter("api:bandwidth_rx", labels! {}).unwrap();
-      assert_eq!(upload.get_counter("api:bandwidth_tx_uncompressed", labels! {}), Some(119));
+      assert!(
+        matches!(
+          upload.get_counter("api:bandwidth_tx_uncompressed", labels! {}),
+          Some(177 | 178)
```
snowp, line 545: how come this is not deterministic?

bd-logger/src/async_log_buffer.rs
```diff
@@ -290,6 +299,23 @@ impl Sender {
...
+  pub fn request_session_id(&self, timeout: StdDuration) -> anyhow::Result<String> {
+    let (response_tx, response_rx) = bd_completion::Sender::new();
+    self.try_send_state_update(StateUpdateMessage::RequestSessionId(response_tx))?;
+
+    response_rx
+      .blocking_recv_with_timeout(timeout)
+      .map_err(|e| anyhow!("failed to receive session ID from async logger: {e}"))?
+  }
+
+  pub fn request_start_new_session(&self, timeout: StdDuration) -> anyhow::Result<()> {
```
snowp, line 314: with this not returning anything do we need to block on a response here?

bd-logger/src/logger.rs
```diff
@@ -475,22 +504,32 @@ impl LoggerHandle {
     self.stats.state_flushing_counters.record(&result);
   }
 
-  #[must_use]
-  pub fn session_id(&self) -> String {
-    self.session_strategy.session_id()
+  pub fn session_id(&self) -> anyhow::Result<String> {
```
snowp, line 507: I'm not sure if having session_id be able to block for up to 5s is a good idea? Given that callers might call this from anywhere this seems a bit risky, given 5s puts us right around the ANR threshold

mattklein123: I can lower the limit. In general this should never happen. Will just make it 1s.

bd-logger/src/async_log_buffer.rs
mattklein123: Yeah I guess we should have a null completion. Will add it.

bd-logger/src/test/logger_integration.rs
mattklein123: don't remember let me check

bd-logger/src/async_log_buffer.rs
mattklein123: Nevermind this does already block on completion?

snowp (approved):

PR #486, opened 2026-05-08
Fix incorrect background state for AppExits

FranAguilera:
## What

Resolves BIT-8093

Fix precedence of log fields over stale global state for previous run logs. This resolves incorrect "foreground" state shown in background app terminations

Also use the report field "running state" to infer the last app state at the moment of termination

## Verification

Further fix and verification at https://github.com/bitdriftlabs/capture-sdk/pull/978

### DIFF
--- a/bd-crash-handler/src/lib.rs
+++ b/bd-crash-handler/src/lib.rs
@@ -266,7 +266,7 @@ impl Monitor {
     ]);
 
     if let Some(running_state) = report.app_metrics().and_then(|app| app.running_state()) {
-      let is_foreground = running_state.starts_with("foreground");
+      let is_foreground = running_state == "foreground";
       fields.insert(
         "foreground".into(),
         if is_foreground { "1" } else { "0" }.into(),
--- a/bd-crash-handler/src/lib.rs
+++ b/bd-crash-handler/src/lib.rs
@@ -265,20 +265,24 @@ impl Monitor {
       ),
     ]);
 
-    if let Some(running_state) = report.app_metrics().and_then(|app| app.running_state()) {
-      let is_foreground = running_state == "foreground";
-      fields.insert(
-        "foreground".into(),
-        if is_foreground { "1" } else { "0" }.into(),
-      );
-    }
-
     if let Some(app_id) = app_id {
       fields.insert("app_id".into(), app_id.into());
     }
 
     match platform {
       Platform::Android => {
+        // Override foreground field using running_state from the FBS report when available.
+        // On Android, running_state maps to ActivityManager.RunningAppProcessInfo importance:
+        // https://developer.android.com/reference/android/app/ActivityManager.RunningAppProcessInfo
+        // Only exact "foreground" importance maps to foreground=1.
+        if let Some(running_state) = report.app_metrics().and_then(|app| app.running_state()) {
+          let is_foreground = running_state == "foreground";
+          fields.insert(
+            "foreground".into(),
+            if is_foreground { "1" } else { "0" }.into(),
+          );
+        }
+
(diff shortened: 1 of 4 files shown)


bd-logger/src/metadata.rs
```diff
@@ -78,10 +82,10 @@ impl MetadataCollector {
     {
       let log_fields = partition_fields(fields);
```
snowp: we don't need to partition anymore at this point since we're just inserting all the logs so we can just `chain(fields.into_iter()..)` directly to save an a few allocations

bd-logger/src/metadata.rs
```diff
@@ -65,6 +65,10 @@ impl MetadataCollector {
   }
 
   /// Creates metadata from log fields without interpolating provider fields
+  /// Returns metadata for logs from the previous run (e.g. crash/exit logs).
+  /// Log OOTB fields take precedence over previous global state so that
+  /// callers can override potentially stale persisted values when a more
+  /// accurate source is available (e.g. foreground state from `ApplicationExitInfo`).
```
snowp:
```suggestion
/// Returns log metadata using the last active global state at the end of the last process run.
/// Log fields take precedence over persisted global state fields to allow the caller to override values in global state, e.g. when the crash handler knows that the event happened in the background. 
/// Does *not* invoke the field providers as these would incorrectly reflect the state of the current process.
```

Something like this maybe?

bd-logger/src/metadata_test.rs
```diff
@@ -301,17 +301,20 @@ fn metadata_from_fields_with_previous_global_state_includes_global_fields() {
...
+  // Log field overrides stale global state.
   assert_eq!(
-    "global_value",
+    "local_value",
     expected_field_value(&metadata.fields, "shared_ootb_key").unwrap()
   );
 
   // shared_custom_key: Custom ("local_value") vs Global ("global_value")
-  // Custom comes AFTER Global in chain, so Custom should win.
+  // Custom field overrides global state.
```
snowp: At this point we can't distinguish between ootb vs custom override so I think we can simplify this test? Since `AnnotatedLogFields` is a map keyed by the field key, we never have conflicting custom vs ootb within the fields map so there is no ordering between ootb and custom

bd-crash-handler/src/lib.rs
```diff
@@ -265,6 +265,14 @@ impl Monitor {
       ),
     ]);
 
+    if let Some(running_state) = report.app_metrics().and_then(|app| app.running_state()) {
```
FranAguilera: Matching android impl as well https://github.com/bitdriftlabs/capture-sdk/pull/978/changes#diff-1dc1c207af0bc13aa494907db2c55d7c46e0f198f29beb9c85c81c9fe9ceef95R176

snowp: Seems like we should make it more obvious in the code how this ties back to all this, maybe move this check into the android block to make it obvious that it's an Android specific string and link to the Android docs to help understand what this refers to? 

Just trying to make it easier for the next person to understand where this magic string is coming from

bd-crash-handler/src/lib.rs
```diff
@@ -265,6 +265,14 @@ impl Monitor {
...
+      let is_foreground = running_state == "foreground";
+      fields.insert(
+        "foreground".into(),
+        if is_foreground { "1" } else { "0" }.into(),
```
snowp: Maybe do a quick test for this as well so we have coverage? Just a simple unit test to test that we correctly extract out the foreground

snowp (approved):
Couple suggestions but lgtm

PR #491, opened 2026-05-14
Expose SDK status

FranAguilera:
## What

Part of BIT-8084

Exposes an snapshopt of the sdk status containing :

```
/// The SDK's initialization state.
pub enum InitializationState {
  /// The SDK has not been started yet.
  NotStarted = 0,
  /// The SDK library has been loaded and the logger is being constructed,
  /// but log processing has not yet begun.
  Loaded     = 1,
  /// The SDK is fully running and processing logs.
  Running    = 2,
  /// The SDK has been force-disabled by the server (e.g., authentication failure).
  Disabled   = 3,
}

/// A point-in-time snapshot of the SDK's operational status.
pub struct SdkStatus {
  /// The current initialization state of the SDK.
  pub initialization_state: InitializationState,

  /// The wall-clock time of the last successful handshake, if any.
  pub last_handshake_time: Option<OffsetDateTime>,

  /// The wall-clock time of the last successful config delivery from the backend, if any.
  pub last_config_delivery_time: Option<OffsetDateTime>,
}
```

## Verification

Verified with sample apps (Android/iOS) in the integration PR https://github.com/bitdriftlabs/capture-sdk/pull/984

### DIFF
--- a/bd-logger/src/client_config.rs
+++ b/bd-logger/src/client_config.rs
@@ -85,6 +85,9 @@ pub struct Config<A: ApplyConfig> {
   // Delegate used to apply the configuration. This allows for easier testing.
   #[allow(clippy::struct_field_names)]
   apply_config: A,
+
+  time_provider: Arc<dyn TimeProvider>,
+  sdk_status_tracker: bd_client_common::sdk_status::SdkStatusTracker,
 }
 
 impl<A: ApplyConfig> Config<A> {
@@ -94,6 +97,7 @@ impl<A: ApplyConfig> Config<A> {
       sdk_directory,
       apply_config,
       Arc::new(bd_time::SystemTimeProvider),
+      bd_client_common::sdk_status::SdkStatusTracker::new(),
     )
   }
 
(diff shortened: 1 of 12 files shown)


bd-logger/src/client_config.rs
```diff
@@ -148,12 +159,20 @@ impl<A: ApplyConfig> Config<A> {
...
-      .await
+      .await;
+
+    if result.is_ok() {
+      self
+        .sdk_status_tracker
+        .record_config_delivery(self.time_provider.now());
+    }
+
+    result
```
mattklein123, line 175: I think this is going to leave the time unset if we use cached config only. In that case I think we want to se the time for when we ACKd the cached config is the most current config via the handshake response.

mattklein123 (changes requested):
LGTM with one small comment.

bd-logger/src/client_config.rs
FranAguilera: good catch, updated

mattklein123: Hmm. Not sure what the right behavior is here. I think you could argue what you have here is correct. You could also argue that what we want is to only update the times when either:
1) we get new config
2) The handshake tells us the config is up to date (See on_handshake_complete where the handshake tells us it is up to date and we mark the config safe).

Thoughts?

FranAguilera: good point, updated for those cases instead

mattklein123 (approved):

PR #492, opened 2026-05-15
Reset exponential backoff after successful handshake

FranAguilera:
Not ready yet, still verifying if this was intentional


FranAguilera:
This is by design <1 min case

PR #494, opened 2026-05-15
[BIT-8126] Add memory state tracking for crash report enrichment

ArielDemarco:
# Overview
Adds the ability to record memory pressure events in the system state, so that crash reports can include the last known memory level at the time of the crash. This warning will include the level (`normal`, `warning` and `critical`) and timestamp of the warning.


bd-crash-handler/src/lib.rs
```diff
@@ -458,6 +458,40 @@ impl Monitor {
...
+      .iter()
+      .filter(|(scope, ..)| *scope == bd_resilient_kv::Scope::System)
+      .filter_map(|(_, name, TimestampedValue { value, .. })| {
+        if !value.has_string_value() {
+          return None;
+        }
+        let field_name: &str = match name.as_str() {
+          "low_memory_level" => "_low_memory_level",
+          "low_memory_used_kb" => "_low_memory_used_kb",
+          "low_memory_timestamp_us" => "_low_memory_timestamp_us",
```
snowp: The state store carries the carries the timestamp of the last time the state was updated so maybe that can be used instead of this if the timestamp is just recording when the event was logged? `TimestampedValue` carries the timestamp

bd-crash-handler/src/lib.rs
```diff
@@ -458,6 +458,40 @@ impl Monitor {
...
+          _ => return None,
+        };
+        Some((field_name.into(), value.string_value().to_string().into()))
+      })
+      .collect();
+
+    // If the last recorded memory pressure level was "normal", memory pressure had resolved
+    // before the crash so we don't include low memory fields in the report
+    if fields
```
snowp: perf nit but if you make this check against the value in the state you avoid constructing new strings for the log fields map that you end up throwing away

bd-logger/src/logger.rs
```diff
@@ -498,6 +498,26 @@ impl LoggerHandle {
     );
   }
 
+  pub fn notify_low_memory(&self, level: String, memory_used_kb: u64, timestamp_us: u64) {
```
snowp: do we want to use a platform agnostic enum for the possible levels? Or are we passing through platform specific values here?

bd-crash-handler/src/lib.rs
ArielDemarco: Good idea; I didn't think about that. I think that the gap between when the memory warning came in and when the metadata was persisted is negligible

bd-crash-handler/src/lib.rs
ArielDemarco: good call

bd-logger/src/logger.rs
ArielDemarco: Its a good question. We still need to define whether this is going to work in a similar way on Android, since the way they handle OOMs is very different from how we do it on iOS.
That said, tbh, seems like a good idea to abstract it into an enum either way

bd-logger/src/async_log_buffer.rs
```diff
@@ -1053,6 +1056,18 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
                     }
                   }
                 },
+                StateUpdateMessage::SetMemoryPressureLevel { level } => {
+                  if let Err(e) = state_store
+                    .insert(
+                      Scope::System,
+                      "memory_pressure_level".to_string(),
+                      int_value(i64::from(level.0)),
```
FranAguilera: instead of reporting as int here can we represent as readable string instead?

here an [example](https://github.com/bitdriftlabs/shared-core/blob/87a0ed82b7c89388da3ba504fb27b40293682cf5/bd-crash-handler/src/lib.rs#L316) for the emitted `_app_exit_reason` to each fatal issue 

Image

```
  fn report_type_to_reason(report_type: ReportType) -> &'static str {
    match report_type {
      ReportType::AppNotResponding => "ANR",
      ReportType::NativeCrash => "Native Crash",
      ReportType::JVMCrash => "Crash",
      ReportType::StrictModeViolation => "Strict Mode Violation",
      ReportType::MemoryTermination => "Memory Termination",
      ReportType::JavaScriptFatalError => "Fatal JavaScript Error",
      ReportType::JavaScriptNonFatalError => "Non-Fatal JavaScript Error",
      ReportType::HandledError => "Handled Error",
      _ => "Unknown",
    }
  }

```

ArielDemarco: Sure; we can

bd-logger/src/logger.rs
ArielDemarco: Solved!

FranAguilera (approved):
Verified with various Android sessions. [Here's one](https://timeline.bitdrift.dev/session/981b9ac0-18d2-4523-aaf6-b57b185c6ca2) with ANR and critical memory warning

bd-logger/src/async_log_buffer.rs
```diff
@@ -1053,6 +1056,20 @@ impl<R: LogReplay + Send + 'static> AsyncLogBuffer<R> {
                     }
                   }
                 },
+                StateUpdateMessage::SetMemoryPressureLevel { level } => {
+                  if let Err(e) = state_store
+                    .insert(
+                      Scope::System,
+                      "memory_pressure_level".to_string(),
```
mattklein123: nit: please use a named constant somewhere for all the places this is duplicated.

bd-logger/src/builder.rs
```diff
@@ -78,6 +79,37 @@ pub fn default_stats_flush_triggers(
   )))
 }
 
+async fn initialize_memory_pressure(
+  previous_run_state: &bd_resilient_kv::ScopedMaps,
+  state_store: &StateStore,
+  previous_memory_pressure_level: &std::sync::atomic::AtomicI8,
+) {
+  let level = previous_run_state
+    .iter()
```
mattklein123: You should be able to use get() here instead of iterating. I saw this elsewhere also.

mattklein123 (changes requested):
LGTM with a few nits

mattklein123 (approved):

PR #495, opened 2026-05-15
client stats: send multiple snapshots in a single request

mattklein123:


bd-client-stats/src/stats.rs
```diff
@@ -55,15 +58,48 @@ use tokio::sync::{mpsc, oneshot, watch};
...
+
+fn batch_transport_uuid(source_file_ids: &[String]) -> String {
+  if source_file_ids.len() == 1 {
+    return source_file_ids[0].clone();
+  }
+
+  // Retries must use the same transport UUID for the same logical batch, but the UUID also needs
+  // to change when the batch composition changes. A stable SHA-256 hex digest of the ordered
+  // source file IDs gives us both properties across retries and restarts, and `upload_uuid` is
+  // only treated as an opaque string in this path.
```
snowp, line 95: This seems fine given we don't really dedupe or anything, but with this I imagine we could have processed payload with snapshot a,b,c then if we retried with a,b,c,d we'd treat it as a new upload and double write? Maybe not worth dealing with now, but maybe this should have N ids in the future so we could apply them incrementally

snowp (approved):

PR #496, opened 2026-05-19
Extract traits for stat collection

kattrali:
Makes a new trait for `StatsCollector` and moves bits related to sdk storage into `Option`s on `WorkflowEngine`. Ideally these would all be encapsulated together, but for now engine has freshly optional properties.


bd-stats-common/src/lib.rs
```diff
@@ -125,3 +126,35 @@ pub trait Counter: Send + Sync + Debug {
...
+    &self,
+    tags: BTreeMap<String, String>,
+    id: &str,
+  ) -> Option<Box<dyn Counter>>;
+
+  fn workflow_dynamic_histogram(
+    &self,
+    tags: BTreeMap<String, String>,
+    id: &str,
+  ) -> Option<Box<dyn Histogram>>;
```
mattklein123, line 159: On the client for perf reasons it would be optimal to not box here and probably just use template traits, but we can look at that as a follow up. This is fine for now.

bd-workflows/src/engine.rs
```diff
@@ -107,21 +107,22 @@ pub struct WorkflowsEngine {
...
     scope: &Scope,
-    sdk_directory: &Path,
-    runtime: &ConfigLoader,
+    sdk_directory: Option<&Path>,
+    runtime: Option<&ConfigLoader>,
     data_upload_tx: Sender<DataUpload>,
-    stats: Arc<Stats>,
-    stats_flush_trigger: FlushTrigger,
+    stats: Arc<dyn StatsCollector>,
+    stats_flush_trigger: Option<FlushTrigger>,
```
mattklein123, line 125: Yeah agreed it would be good to clean this up a bit more but we can look at this in a follow up.

mattklein123 (approved):

PR #517, opened 2026-06-09
Allow CRASH_LOOP_BYPASS_TIMEOUT_SECONDS to be configurable

FranAguilera:
Resolves BIT-8645

Allowing to configured CRASH_LOOP_BYPASS_TIMEOUT_SECONDS via shard config flag `safe_file_cache.crash_loop_bypass_timeout_ms`

Verified with `gradle-test-app`. See session [here](https://timeline.bitdrift.dev/session/ab56709c-ad0b-4d63-9872-8e3d5a6f44ea?utm_source=sdk)


FranAguilera:
Current impl is off, as will wipe the config during crash loop detection

PR #520, opened 2026-06-11
buffers: add durable persistence to trigger uploads and streaming

mattklein123:
This is a very large and risky change that adds durable persistence to buffer uploads. Previously there were many cases where uploads could be halted after restarts, or portions of uploads could be lost after restart. Also explicit flush buffers did not previously support streaming and now they do.

Fixes BIT-8318


bd-buffer/src/buffer/mod.rs
```diff
@@ -159,8 +159,9 @@ pub trait RingBufferCursorConsumer: Send {
   // practice, this API is paired with startRead() and can only be called after a successful call to
   // startRead(), but does not have to interleave back and forth with that API. I.e., it is possible
   // to call startRead(), startRead(), advanceReadPointer(), startRead(), etc. Calling this releases
-  // the oldest read data acquired with startRead().
-  fn advance_read_pointer(&mut self) -> Result<()>;
+  // the oldest read data acquired with startRead(). Count specifies how many records to advance the
+  // read pointer forward.
```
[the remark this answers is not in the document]
snowp: seems legit? minor comment

bd-buffer/src/buffer/non_volatile_ring_buffer.rs
```diff
@@ -581,23 +581,30 @@ impl RingBufferCursorConsumer for CursorConsumerImpl {
     };
 
     let mut common_ring_buffer = parent.common_ring_buffer.locked_data.lock();
+    if count == 0 {
+      return Ok(());
+    }
```
snowp, line 585: I'm sure this is an unusual edge case but seems like we should do this before locking and maybe even before upgrading?

bd-logger/src/client_config.rs
```diff
@@ -285,6 +285,12 @@ impl ApplyConfig for LoggerUpdate {
       filters,
     } = configuration;
 
+    // This await is the ordering barrier for trigger-upload recovery. The buffer manager does not
```
snowp: Maybe word this to make it clear that all of this is for sequencing of the first config update? ApplyConfig is called on each update so this logic is only relevant for the startup case

All this sequencing stuff is scary but makes sense

bd-logger/src/log_replay.rs
```diff
@@ -606,10 +635,30 @@ impl ProcessingPipeline {
...
+          remote_flush_streaming.buffer_ids.into_iter().collect(),
+          &remote_flush_streaming.session_id,
+          remote_flush_streaming.streaming,
+        );
+
+        if started {
+          self
+            .is_tracing_active
+            .store(self.workflows_engine.is_tracing_active(), Ordering::Relaxed);
+        }
```
snowp, line 660: why is this bit necessary? wouldn't we have started tracing already if we were tracing?

bd-logger/src/trigger_upload_artifact.rs
```diff
@@ -0,0 +1,256 @@
...
+
+const TRIGGER_UPLOAD_ARTIFACTS_DIRECTORY: &str = "trigger_upload_artifacts";
+
+//
+// PersistedTriggerUploadArtifactBatch
+//
+
+// Concrete log batch that has been durably staged for a single trigger upload / buffer pair.
+// `upload_uuid` is persisted with the logs so retries and restart recovery keep using the same
+// upload identity instead of minting a fresh UUID for the same payload.
```
snowp, line 33: Seems like we should be doing the same for continuous buffer uploads to keep the uuid between restarts?

bd-logger/src/flush_registry.rs
```diff
@@ -0,0 +1,721 @@
...
+            {
+              buffer.lifecycle = PersistedTriggerUploadBufferLifecycle::UploadingFromBuffer;
+            }
+          }
+        }
+      })
+      .await;
+  }
+
+  pub async fn mark_uploading_from_artifact(&self, id: &str, buffer_id: &str) {
```
snowp, line 545: How come we need to track whether we are uploading from artifact or buffer explicitly like this, seems like the presence of the artifact file could be enough make that call on resumption? Or is this just to keep all the state tracking centralized?

snowp (commented):
Some questions but makes sense to me. Probably too complicated to try to runtime guard this, if not we'll just have to test well

snowp (approved):

bd-buffer/src/buffer/mod.rs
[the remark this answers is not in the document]
mattklein123: Fixed in 9c8898cc. I updated the example to `advance_read_pointers(1)` so it matches the renamed API and signature.

bd-logger/src/flush_registry.rs
```diff
@@ -0,0 +1,746 @@
...
+      Err(e) => {
+        // A corrupt snapshot is treated as unrecoverable stale state. We delete it and proceed
+        // empty rather than trapping the logger in a permanent startup failure loop.
+        log::warn!(
+          "failed to deserialize pending trigger uploads from {}: {e}",
+          state_path.display()
+        );
+        let _ignored = delete_file_if_exists_async(state_path).await;
+        Vec::new()
+      },
```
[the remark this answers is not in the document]
mattklein123: Fixed in 9c8898cc. The corrupt-snapshot path now removes the `trigger_upload_artifacts` directory as well, and I added coverage in `corrupted_snapshot_is_dropped_and_treated_as_empty` to verify both the snapshot and orphaned artifacts are cleaned up.

bd-buffer/src/buffer/non_volatile_ring_buffer.rs
mattklein123: Updated in 9c8898cc to return early before taking the lock. I left the `upgrade()` first so `advance_read_pointers(0)` still preserves the existing shutdown behavior instead of silently succeeding after the consumer is gone.

bd-logger/src/client_config.rs
mattklein123: Reworded in 9c8898cc to call out that this sequencing only matters for startup / the first trigger-buffer config update.

bd-logger/src/log_replay.rs
mattklein123: This is needed because remote flush streaming can start after the initial `workflows_engine.start(...)` path. `start_remote_flush_streaming(...)` can activate a tracing lease, so the atomic `is_tracing_active` mirror has to be refreshed after a successful start; otherwise the logger-side fast path would keep seeing the old value until some later config-driven refresh.

bd-logger/src/trigger_upload_artifact.rs
mattklein123: I left this trigger-upload-only for now. The stable UUID here is specifically for replaying the same durably staged trigger-upload batch across restart/retry; continuous uploads do not currently persist a per-batch artifact backlog, so after restart they build a new batch rather than resuming the identical payload. If we later add durable continuous-batch replay, stable UUID reuse would make sense there too, but I did not want to broaden this PR.

bd-logger/src/flush_registry.rs
mattklein123: The explicit lifecycle is intentional. Artifact-file presence only tells us that some staged backlog exists; it does not tell us whether a given upload/buffer is still reading live buffer contents, has switched to artifact replay, or has already recorded durable per-buffer progress within that backlog. We use the registry as the authoritative restart-safe state, so keeping that phase there avoids inferring recovery behavior from partial filesystem state alone.

bd-logger/src/trigger_upload_artifact.rs
mattklein123: (non-ai answer, we should do this, but this PR is too big).

PR #536, opened 2026-06-26
session strategy: run callbacks on calling thread

mattklein123:
There are regressions with calling the callbacks cross thread which are causing deadlocks. This brings us back to the previous behavior where we run the calbacks on the calling thread and run the async persist work on the background thread.


bd-logger/src/logger.rs
```diff
@@ -530,9 +530,13 @@ impl LoggerHandle {
     let is_allowed = LOGGER_GUARD.with(|cell| cell.try_borrow().is_ok());
 
     if is_allowed {
-      self
+      let prepared = self.session_strategy.prepare_session_id()?;
+      let session_id = prepared.current_session_id().to_string();
+      let callback = self
         .tx
-        .request_session_id(async_log_buffer::SESSION_BRIDGE_TIMEOUT)
+        .persist_prepared_session(prepared, async_log_buffer::SESSION_BRIDGE_TIMEOUT)?;
```
mattklein123, line 542: We don't have to do this unless something changed.

bd-session/src/lib.rs
```diff
@@ -181,53 +239,77 @@ impl Strategy {
...
+
+  pub async fn session_id(&self) -> anyhow::Result<String> {
+    let prepared = self.prepare_session_id()?;
+    let session_id = prepared.current_session_id().to_string();
+    let callback = self.persist_prepared(prepared).await;
+    self.run_prepared_callback(callback);
+    Ok(session_id)
+  }
+
+  pub async fn start_new_session(&self) {
```
mattklein123: Remove this?

FranAguilera (approved):

PR #544, opened 2026-07-08
[iOS] Reduce fbs size

FranAguilera:
## Goal

Part of BIT-8854

Similar to https://github.com/bitdriftlabs/shared-core/pull/541, https://github.com/bitdriftlabs/shared-core/pull/543 and https://github.com/bitdriftlabs/capture-sdk/pull/1031

Applying memoization for the generated fbs report on iOS. 

This will guarded in https://github.com/bitdriftlabs/capture-sdk/pull/1052 with a similar killswitch as done in [Android](https://github.com/bitdriftlabs/capture-sdk/pull/1031) 

## Killswitch

`client_feature.ios.optimize_fatal_issue_report_size`, enabled by default. Added in https://github.com/bitdriftlabs/capture-sdk/pull/1052

### DIFF
--- a/bd-report-writer-tests/src/ffi_tests.c
+++ b/bd-report-writer-tests/src/ffi_tests.c
@@ -48,7 +48,7 @@ const uint8_t *load_full_report(BDProcessorHandle handle, uint64_t *len) {
 }
 
 void create_handle(BDProcessorHandle handle) {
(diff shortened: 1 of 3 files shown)


ArielDemarco (approved):
LGTM

PR #552, opened 2026-07-16
[BIT-9154] bd crash reporter killswitch

ArielDemarco:


bd-crash-handler/src/config_writer.rs
```diff
@@ -57,27 +59,34 @@ impl ConfigWriter {
...
         _ = self.crash_reporting_enabled_flag.changed() => {},
+        _ = self.use_bd_crash_reporter_flag.changed() => {},
         () = self.shutdown.cancelled() => return,
       };
 
       // There is chance here of the value changing during platform read, in
       // which case crash reporting would be disabled for that session
 
       let _ = self.crash_reporting_enabled_flag.read_mark_update();
+      let _ = self.use_bd_crash_reporter_flag.read_mark_update();
```
snowp, line 70: I don't really know why we don't just do read_mark_delete in the function below but seems fine to not mess with this

bd-crash-handler/src/config_writer.rs
```diff
@@ -57,27 +59,34 @@ impl ConfigWriter {
...
+       report config file {}",
       self.config_path.display()
     );
 
     self.try_ensure_directories_exist().await;
 
-    let contents = format!("crash_reporting.enabled,{crash_reporting_enabled}");
+    let contents = format!(
+      "crash_reporting.enabled,{crash_reporting_enabled}\nclient_feature.ios.\
+       use_bd_crash_reporter,{use_bd_crash_reporter}"
```
snowp, line 88: Let's make sure all the current platform parsers handle this, we don't want a sdk downgrade to blow up

bd-runtime/src/runtime.rs
```diff
@@ -662,6 +662,14 @@ pub mod crash_reporting {
   // other fatal events. This defaults to enabled so fresh installs still monitor
   // crashes until runtime explicitly turns it off.
   bool_feature_flag!(Enabled, "crash_reporting.enabled", true);
+
+  // Enables or disables the use of the bitdrift in-process crash reporter on iOS,
+  // as opposed to relying solely on KSCrash/MetricKit.
+  bool_feature_flag!(
+    UseBdCrashReporter,
+    "client_feature.ios.use_bd_crash_reporter",
+    true
```
snowp, line 671: I know this is how we did it for crash workflows but maybe double check that the csv is written early enough for the platform layer to see it? Or if we're handling missing in the platform layer that is also ok

snowp (approved):
lgtm lets just verify a few things

bd-crash-handler/src/config_writer.rs
ArielDemarco: yep, tbh, i just kept it consistent with the existing pattern

bd-crash-handler/src/config_writer.rs
ArielDemarco: sure; checked both platform parsers:
- iOS: splits per-line, comma-delimited, keyed generically so unrecognized keys are just extra map entries.
- Android: does the exact same line-based generic parse.

Neither one special-cases the number or set of keys, so an older SDK reading a newer two-line `config.csv` (or a newer SDK reading an older one-line file) parses fine either way

bd-runtime/src/runtime.rs
ArielDemarco: Yes, i manually tested this... and shares the same timing model as the existing crash flag: on first install/clear-cache the file doesn't exist yet, and the platform layer already treats a missing file as "use defaults". I extended that same fallback to the new key, so if the new key is absent from the file (missing entirely, or present in an old-format file that predates this flag), the iOS platform layer falls back to the flag's default instead of failing

PR #564, opened 2026-08-03
multiple client stats changes

mattklein123:
1) Implement direct stats sending in the handshake 
2) Implement per-process mux connection count
3) Implement per-handshake stats analytics


bd-client-stats/src/lib.rs
```diff
@@ -151,6 +153,12 @@ impl Stats {
   ) -> FlushHandles {
     let flush_time_histogram = self.collector.scope("stats").histogram("flush_time");
     let (flush_trigger, flush_rx) = FlushTrigger::new();
+    // HandshakeStats is owned by the API task while Flusher runs in a sibling task. Each
+    // API-originated upload creates a fresh tracked request whose sender belongs to the stream
+    // StateTracker, but its receiver must be consumed by Flusher to complete or abandon claimed
+    // files. This handoff is dynamic because reconnects create new StateTrackers and batches.
+    let (api_upload_completion_tx, api_upload_completion_rx) =
+      tokio::sync::mpsc::unbounded_channel();
```
snowp: Do we want this to be bounded? Not sure I see a reason why we can't put some limit on this?

bd-client-stats/src/stats.rs
```diff
@@ -69,17 +72,127 @@ pub enum PeriodicAction {
...
+      source_file_ids,
+    }) = self.file_manager.get_or_create_pending_upload(true).await?
+    else {
+      return Ok(None);
+    };
+
+    let upload_uuid = batch_transport_uuid(&source_file_ids);
+    request.upload_uuid.clone_from(&upload_uuid);
+    request.sent_at = self.time_provider.now().into_proto();
+    request.upload_reason = UploadReason::UPLOAD_REASON_PERIODIC.into();
```
snowp: Should there be a new reason for this handshake one? You can obviously tell that it's coming in through the handshake just seems strange to call it periodic

snowp (approved):

PR #568, opened 2026-08-05
add plan for new event buffer design

snowp:


.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+## Control-flow ownership
+
+EventBuffer is the ordered data-plane ingress, not a general control bus. The migration retains
+the following ownership boundaries.
+
+| Flow | Owner and transport | Why it does or does not enter EventBuffer |
+| --- | --- | --- |
+| Logs, feature-flag exposure, post-startup memory pressure/entity-ID persistence, and `FlushState` | EventBuffer entries | These are ordered workflow, state-store, or barrier inputs. `FlushState` and `Block::Yes` logs are protected entries. |
+| Logger `setField` / `removeField` | Synchronous EventBuffer-owned COW map | These mutate admission metadata rather than producing a replayable workflow input. |
+| Session creation/rotation, durable persistence, and session flush | `bd_session` generation-based coalescing flusher | Session mutation remains under the strategy's existing mutex. A single flusher persists only the latest generation and has no per-operation FIFO command queue. EventBuffer receives only the resolved immutable session ID. |
```
mattklein123: Any mutex that may be held on the caller thread should use the OS preferred mutex in the end, including this one for sure. I would update the plan to reflect that.

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+
+Each admission receives a monotonic insertion ID and is appended directly to `VecDeque<Entry>`.
+On pressure, EventBuffer scans the deque to select strictly lower-priority victims. It evicts
+lowest-priority candidates first and, within a victim priority, newer entries first; this preserves
+the oldest retained entry for an equal-priority admission.
+
+- Eviction physically removes selected entries with `VecDeque::remove`, in descending index order
+  so earlier selected indexes remain valid. There are no tombstones, stale index records, or
+  compaction passes in this version. Contention benchmarks measure lock hold time for this linear
+  scan/removal strategy before any index or compaction design is considered.
```
mattklein123: I'm not convinced this is the right data structure for this. We will need to benchmark and think about it. Removing from the middle of a Vec<Deque> is not fast. I would either:
1. Use a LinkedList for this
2. Potentially use N VecDequeue, 1 for each different priority/classification level.

Personally I think N Vec<Dequeue> is likely the best option.

If we go with N dequeues we would still need to keep track of sequence for each and I guess drop ranges so that we can make sense of where to pull from? This all needs more thinking.

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+Session durability remains outside EventBuffer. `bd_session` keeps ownership of its existing
+mutex-protected in-memory state; EventBuffer never holds a session strategy handle, persistence
+entry, callback, or session-persistence byte charge.
+
+Milestone 1 replaces snapshot-carrying `PreparedSessionOperation` persistence with a
+generation-based coalescing flusher inside `bd_session`:
+
+- A session mutation applies under the existing strategy mutex, increments a monotonic dirty
+  generation, and returns the new/current session ID immediately. It records any deferred callback
+  against that generation, rather than handing a cloned `LoadedState` to a logger queue.
```
mattklein123: I don't really understand the implications of what happens if the process ends before the updated session ID is made durable. Maybe it's fine? It seems like we would implicitly end up starting a new session on restart. I would explore these edge cases before you finalize the plan.

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+   rotation records a deferred callback for post-admission dispatch without waiting for persistence.
+   The caller then holds
+   `with_thread_local_logger_guard` and captures provider timestamp and fields outside the
+   EventBuffer lock. A provider failure is also a terminal drop; both outcomes record their
+   respective metrics and resolve any `Block::Yes` completion without entering EventBuffer.
+2. `PreviousRunSessionID` logs skip current-process session, provider, and logger-field capture.
+   They retain their raw fields and override for the existing previous-global-state consumer path.
+   Normal logs and `OccurredAt` logs proceed with their captured provider data; the latter retains
+   its supplied occurrence timestamp.
+3. EventBuffer acquires its lock and snapshots the COW logger-field map. The log entry retains the
```
mattklein123: Is this really COW? Do you mean a mutex protected Arc?

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+   snapshot, field-map snapshot, session ID, and optional completion handle. A rejected log does
+   not roll back or otherwise alter an already-persisted session transition.
+4. Admission applies the total/log limits and priority eviction policy. A `Block::Yes` log is
+   protected, so it bypasses `log_limit` and cannot be evicted; it is rejected only if it cannot
+   fit the remaining `total_limit` after evictable entries have been displaced. Rejection or
+   eviction resolves the entry's completion with a terminal drop outcome after releasing the lock.
+   Successful admission schedules `Notify`; it does not wait for the background consumer. In
+   either terminal source outcome, dispatch any deferred implicit-session callback only after the
+   lock is released, so callback-originated logging follows this source operation.
+5. The consumer removes the entry in FIFO order, runs the existing interceptors, then normalizes
```
mattklein123: Per above we need to think more carefully how the buffer will work, especially if there are multiple queues.

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+Removing `PreConfigBuffer` at this point means startup events are retained in their original
+EventBuffer representation, so the same priority/eviction policy applies before and after
+configuration is ready.
+
+### Previous-session replay ordering
+
+`CapturedLog` carries two independent classifications: its retention priority and its source
+(`CurrentProcess` or `PreviousProcess`). A previous-process log is eligible for special ordering
+only when EventBuffer admits it while the startup gate is holding. The eligibility bit is captured
+on the entry; it is not inferred later from the source alone.
```
mattklein123, line 558: How long is the startup gate by default? How is it configured?

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+The same lock transition changes the gate to `Open`. Entries admitted afterwards always append to
+the back, including a late previous-process crash log. The gate never reopens, so a crash report
+that arrives later in the session keeps its higher retention priority but is not reordered ahead of
+current-session work. This avoids retroactively changing workflow order after current-session
+events have started flowing.
+
+Only prior-process logs move during this partition. Current state/control entries, fields,
+session-bearing entries, and flush barriers stay in their original FIFO relation. The one-time
+partition moves entries without cloning payloads; it is an intentional startup-only lock hold and
+is instrumented separately.
```
mattklein123: Seems like with multiple queues this all gets a lot simpler.

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+
+An admitted `FlushState(Block::Yes)` or blocking log is also a gate barrier: it seals the gate,
+partitions eligible previous-process logs first, then drains through its ordered position before
+its completion resolves. It does not bypass older work. An admission-rejected or
+provider-capture-rejected blocking operation resolves immediately as a terminal drop and cannot
+act as a barrier. `FlushState(Block::No)` stays behind the gate, matching its existing
+fire-and-forget behavior. Neither admitted blocking operation may remain pending solely because
+the soft startup delay has not elapsed.
+
+## Observability and validation
```
mattklein123, line 589: Not clear to me that we should ship all of these metrics. We need to think through what is useful for local benchmarking vs. wide production rollout.

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,553 @@
...
+- It deliberately accepts that provider execution can add synchronous caller latency, subject to
+  measured provider and lock-tail latency.
+- Session persistence is deliberately outside EventBuffer. `bd_session` applies a session mutation
+  immediately under its own mutex and coalesces durable writes by generation; logs and state
+  operations carry the resulting immutable session ID. EventBuffer admission never changes,
+  persists, or rolls back session state.
+- EventBuffer is one ordering domain for producer data and state, not for configuration. Existing
+  configuration control-plane timing remains out of band and is made explicit through the startup
+  gate.
+- The initial limits are an 11 MiB total budget and a 1 MiB evictable-log budget, preserving the
```
mattklein123: Is this right? Or you mean this includes disk like today. 11 seems high not sure where this is coming from.

.plans/event-buffer-plan.md
```diff
@@ -0,0 +1,844 @@
...
+with logs: log entries, selected state operations, and flush barriers. It replaces the current log
+and state ingress channels while retaining one async consumer.
+
+- The producer path only captures inputs and admits an entry. It never waits for disk I/O, provider
+  calls while holding the buffer lock, or downstream workflow processing.
+- Five fixed retention tiers decide what is discarded under memory pressure; admission IDs preserve
+  global FIFO delivery across those tiers. Previous-process logs have a temporary startup lane so
+  they can replay before current-process work without a drain-time queue partition.
+- `bd_session` remains responsible for session state and persistence. EventBuffer uses a small,
+  protected fence to wait on the existing session coalescer before a still-current session's work
```
snowp: Currently this includes a fence that avoids processing logs before their session has been persisted. I'm not sure if this is necessary but would be another guarantee that we can try to enforce. This is one of the things I'd want to decide with others on whether just best effort persistence is OK

.plans/event-buffer-plan.md
```diff
@@ -1,12 +1,51 @@
...
+  reaches downstream durable buffers. A newer session supersedes an older fence rather than
+  writing stale session state.
+- The buffer begins with a 10 MiB total in-memory limit and a 1 MiB ordinary-log sublimit. Both
```
mattklein123: Isn't 10MiB much bigger than today? Where does this come from?

mattklein123 (approved):

PR #572, opened 2026-08-05
Support Nested matching via JsonPathValue/logFieldJsonPath

FranAguilera:
Adds support for nested json matching via the existing `Leaf::JsonPathValue`, for this initial prototype I'm proposing to use existing JSON string to extract the nested conditions

More details, regarding frotnend/rollout/performance etc in the [internal PRD here](https://docs.google.com/document/d/1Ln5ecQfEbaX-sOLY7-l9bUwTbUHkSFsT56ncmGgFhwc/edit?tab=t.0) 

You can run local benchmark tests with `make benchmark-json-matcher`

This is used in https://github.com/bitdriftlabs/capture-sdk/pull/1079

This new logic will be controlled by killswitch `workflows.json_path_string_matching_enabled` (which is enabled by default)


bd-log-matcher/src/matcher.rs
```diff
@@ -653,6 +665,28 @@ fn parse_json_path(key_or_index: &KeyOrIndex) -> Result<JsonPathToken> {
 }
 
 fn resolve_json_path<'a>(value: &'a DataValue, path: &[JsonPathToken]) -> Option<Cow<'a, str>> {
+  // Existing SDK APIs send JSON as a string. Parsing is only reached from JsonPathValue matchers.
+  // TODO: Gate JSON string extraction with a remote workflow runtime flag.
+  if let Some(json) = value.as_str() {
+    return resolve_json_string_path(json, path);
```
FranAguilera: Pending to add a killswitch flag here so we can disable remotely in case something is off

bd-log-matcher/src/json_path.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+#[derive(Clone, Copy)]
+struct StringToken {
+  start: usize,
+  end: usize,
+  escaped: bool,
+}
+
+enum PathResult<'a> {
```
snowp: I think this is just Option<Cow<'a, str>>, I don't think the type adds that much and prevents ? usage

Looking at how we do return Option<PathResult<_>> I'd either model it as 

```
enum PathResult {
  Found(Cow<'a, str>,
  Missing,
  Invalid
}
```

if we want to understand the tri-state or just collapse it to an Option<Cow<'a, str>> for simplicity

bd-log-matcher/src/json_string_matcher_test.rs
```diff
@@ -0,0 +1,119 @@
...
+fn matches_json_string_boolean_and_number_values() {
+  let enabled = json_string_matcher(vec![key("enabled")], "true");
+  let count = json_string_matcher(vec![key("count")], "42");
+
+  assert!(matches(&enabled, r#"{"enabled":true}"#));
+  assert!(matches(&count, r#"{"count":42}"#));
+}
+
+#[test]
+#[ignore = "run manually with --release -- --ignored --nocapture"]
```
snowp: This would be better off as a criterion test - I just added one that we can use for reference. It handles the standard stuff like running it multiple times, percentiles etc. Not a blocker for this PR but it should make the benchmarking experience better

bd-log-matcher/src/json_path.rs
```diff
@@ -0,0 +1,480 @@
...
+  Found(Cow<'a, str>),
+  Missing,
+}
+
+impl<'a> Parser<'a> {
+  const fn new(input: &'a str) -> Self {
+    Self { input, pos: 0 }
+  }
+
+  fn walk_value(&mut self, path: &[JsonPathToken]) -> Option<PathResult<'a>> {
```
snowp: With these kind of recursive parsers one thing we have to be careful about when we don't control the input is stack overflow. The simplest thing to do is to add a depth limit (say 128) to prevent going to deep, or switch over to an iterative approach that doesn't have the same per-depth overhead. I think a depth limit is reasonable so we can probably start there unless we believe someone really cares about super complex JSON objects

bd-log-matcher/src/json_path.rs
```diff
@@ -0,0 +1,480 @@
...
+  match byte {
+    b'0' ..= b'9' => Some(byte - b'0'),
+    b'a' ..= b'f' => Some(byte - b'a' + 10),
+    b'A' ..= b'F' => Some(byte - b'A' + 10),
+    _ => None,
+  }
+}
+
+#[cfg(test)]
+mod tests {
```
snowp: Per the guidelines these would go in json_path_test.rs, try out https://github.com/bitdriftlabs/ai-instructions if you need to set up your agent and it should do this automatically

snowp: I'd probably add some more tests for correctness, there's a lot of test surface. The agent will do a good job here

fuzz/src/json_path.rs
```diff
@@ -0,0 +1,44 @@
...
+use bd_log_matcher::matcher::{JsonPathToken, fuzz_json_path};
+
+#[derive(Debug)]
+pub struct JsonPathFuzzTestCase {
+  json: Vec<u8>,
+  path: Vec<JsonPathToken>,
+}
+
+impl<'a> Arbitrary<'a> for JsonPathFuzzTestCase {
+  fn arbitrary(input: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
```
snowp, line 22: One thing you can do here to validate correctness via fuzzing is to separately implement this via serde_json - the implementation is much simpler and you get to validate that our more complex custom parser agrees with the simple implementation. Then you can let it run and catch any issues (or have the agent do it)

Love using a fuzzer for this!

bd-log-matcher/src/matcher.rs
```diff
@@ -653,6 +665,28 @@ fn parse_json_path(key_or_index: &KeyOrIndex) -> Result<JsonPathToken> {
 }
 
 fn resolve_json_path<'a>(value: &'a DataValue, path: &[JsonPathToken]) -> Option<Cow<'a, str>> {
+  // Existing SDK APIs send JSON as a string. Parsing is only reached from JsonPathValue matchers.
+  // TODO: Gate JSON string extraction with a remote workflow runtime flag.
+  if let Some(json) = value.as_str() {
```
snowp: just for code clarity I would do something like `if let Some(maybe_json) ...` as I think it more clearly captures that we aren't expecting JSON here but are opportunistically trying to extract a JSON value per the matcher

fuzz/src/json_path.rs
snowp, line 22: I don't know if serde_json supports arbitrary out of the box but it would be interesting to have the agent 1) generate sensible JSON and 2) try to ensure that occasionally they match. If you look at the state fuzzer we do stuff like that to try to guide it to sane values

bd-log-matcher/src/json_path.rs
```diff
@@ -0,0 +1,480 @@
...
+
+pub(super) fn resolve<'a>(input: &'a str, path: &[JsonPathToken]) -> Option<Cow<'a, str>> {
+  let mut parser = Parser::new(input);
+  match parser.walk_value(path)? {
+    PathResult::Found(value) => Some(value),
+    PathResult::Missing => None,
+  }
+}
+
+struct Parser<'a> {
```
snowp, line 21: I would add some docs on the perf implications of this, there's a few different things worth calling out when it needs to allocate vs not e.g. for escaped values

snowp (commented):
I think we need a depth bound on the parser to deal with bad input but other than that all my comments are just suggestions or style guidance, good job!

bd-log-matcher/src/json_path.rs
FranAguilera: great idea, will add

FranAguilera: Added 128 limit

bd-log-matcher/src/matcher.rs
FranAguilera: Added flag `workflows.json_path_string_matching_enabled`

bd-log-matcher/src/json_string_matcher_test.rs
```diff
@@ -0,0 +1,144 @@
...
+use bd_log_primitives::tiny_set::TinyMap;
+use bd_log_primitives::{DataValue, EMPTY_FIELDS, FieldsRef, LogFields, LogMessage, log_level};
+use bd_proto::protos::log_matcher::log_matcher::{LogMatcher, log_matcher};
+use bd_proto::protos::logging::payload::LogType;
+use bd_proto::protos::value_matcher::value_matcher::json_path_value_match::{
+  KeyOrIndex,
+  key_or_index,
+};
+use bd_proto::protos::value_matcher::value_matcher::{JsonPathValueMatch, Operator};
+fn json_string_matcher(path: Vec<KeyOrIndex>, value: &str) -> Tree {
```
snowp, line 19: nit: newline to break the import section from functions

bd-log-matcher/src/matcher.rs
snowp, line 192: Seems safer to force the caller to provide a MatchContext, seems easy to try to call this with the default. I'd add helpers in test to call this with the default over adding the extra function

bd-log-matcher/src/matcher.rs
```diff
@@ -231,44 +316,83 @@ impl Tree {
...
           matcher,
-        } => fields
-          .field(field_key)
-          .and_then(|value| resolve_json_path(value, path))
-          .is_some_and(|input| matcher.evaluate(input.as_ref(), extracted_fields)),
+        } => {
+          let Some(value) = fields.field(field_key) else {
+            return MatchResult::NotMatched;
+          };
+          if !context.json_path_string_matching_enabled && value.as_str().is_some() {
```
snowp, line 302: I think we should plan for this to be removed at some point, it's adding quite a lot of code complexity

bd-log-matcher/src/matcher.rs
```diff
@@ -652,7 +785,32 @@ fn parse_json_path(key_or_index: &KeyOrIndex) -> Result<JsonPathToken> {
...
 }
 
-fn resolve_json_path<'a>(value: &'a DataValue, path: &[JsonPathToken]) -> Option<Cow<'a, str>> {
+fn resolve_json_path<'a>(
+  value: &'a DataValue,
+  path: &[JsonPathToken],
+  json_string_matching_enabled: bool,
+) -> Option<Cow<'a, str>> {
+  // Existing SDK APIs send JSON as a string. Parsing is only reached from JsonPathValue matchers.
+  if json_string_matching_enabled && let Some(json) = value.as_str() {
```
snowp: Seems like we already gate accessing this function on this flag? Don't think we need it in both locations

bd-workflows/src/engine_state_change_test.rs
```diff
@@ -615,6 +624,120 @@ async fn state_change_with_extra_state_matcher_no_match() {
   );
 }
 
+#[tokio::test]
+async fn state_change_extra_json_matcher_respects_runtime_flag() {
+  let b = state("B");
+  let json_matcher = LogMatcher {
```
snowp: nit would be nice with some helpers to avoid this huge inline config

snowp (approved):
Nice, just a few smaller comments but lgtm overall

bd-log-matcher/src/matcher.rs
FranAguilera: I agree, will add a tracking task

FranAguilera:
Re-verified locally

PR #574, opened 2026-08-06
[Android] [ANR] Send basic reports without ANR traces

FranAguilera:
## Problem

[ApplicationExitInfo](https://developer.android.com/reference/android/app/ApplicationExitInfo) can return a null stacktrace sometimes via [getTraceInputStream](https://developer.android.com/reference/android/app/ApplicationExitInfo#getTraceInputStream()) parameter given this definition on their docs

_“Note that because these traces are kept in a separate global circular buffer, crashes may be overwritten by newer crashes (including from other applications), so this may still return null.”_

If that’s the case, getPreviousRunInfo will return ANR as termination reason, but given no report is stored due to the lack of stactrace the count won’t increment in Issue pages/Callback/ANR workflows

For more context please refer to this [doc](https://docs.google.com/document/d/1PHxyRpKdAkza3LC_e0mxkXiDDbO_hJBxIbxo_Lhfy-8/edit?tab=t.0)

## Solution

Resolves BIT-9198

If we fail to parse a trace (either because being nullable or invalid contents) instead of dropping the report will create a basic one empty stacks/threads using appExitDescription for classification (e.g. User Perceived ANR, BroadcastReceiver, Background ANR, etc).

## Verification

Verified in capture-sdk shared-core bump pr https://github.com/bitdriftlabs/capture-sdk/pull/1083

Screenshot 2026-08-06 at 15 48 44

### DIFF
--- a/bd-report-parsers/src/android.rs
+++ b/bd-report-parsers/src/android.rs
@@ -54,11 +54,35 @@ type FrameOffsetSequence = Vec<u32>;
 
 const MAIN_THREAD: &str = "main";
 
-pub fn build_anr<'a, 'fbb>(
+pub fn build_anr_from_app_exit<'fbb>(
   builder: &mut FlatBufferBuilder<'fbb>,
   app_info: &mut v_1::AppMetricsArgs<'fbb>,
   device_info: &mut v_1::DeviceMetricsArgs<'fbb>,
-  input: MemmapView<'a>,
+  anr_trace: Option<MemmapView<'_>>,
+  app_exit_description: Option<&str>,
+  is_file_size_optimization_enabled: bool,
+) -> WIPOffset<v_1::Report<'fbb>> {
+  if let Some(anr_trace) = anr_trace
+    && let Ok((_, report)) = build_anr_from_trace(
+      builder,
+      app_info,
+      device_info,
+      anr_trace,
+      app_exit_description,
+      is_file_size_optimization_enabled,
+    )
+  {
+    return report;
+  }
+
+  build_anr_without_trace(builder, app_info, device_info, app_exit_description)
+}
+
+fn build_anr_from_trace<'a, 'fbb>(
+  builder: &mut FlatBufferBuilder<'fbb>,
+  app_info: &mut v_1::AppMetricsArgs<'fbb>,
+  device_info: &mut v_1::DeviceMetricsArgs<'fbb>,
+  anr_trace: MemmapView<'a>,
   app_exit_description: Option<&str>,
   is_file_size_optimization_enabled: bool,
 ) -> IResult<MemmapView<'a>, WIPOffset<v_1::Report<'fbb>>, nom::error::Error<MemmapView<'a>>> {
@@ -69,7 +93,7 @@ pub fn build_anr<'a, 'fbb>(
     thread_counter,
     |text| build_threads(builder, text, is_file_size_optimization_enabled),
   )
-    .parse(input)?;
+    .parse(anr_trace)?;
 
   if let Ok(pid) = u32::try_from(pid) {
     app_info.process_id = pid;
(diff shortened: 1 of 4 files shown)


bd-report-parsers/src/android.rs
```diff
@@ -54,11 +54,35 @@ type FrameOffsetSequence = Vec<u32>;
...
+pub fn build_anr_from_app_exit<'a, 'fbb>(
   builder: &mut FlatBufferBuilder<'fbb>,
   app_info: &mut v_1::AppMetricsArgs<'fbb>,
   device_info: &mut v_1::DeviceMetricsArgs<'fbb>,
-  input: MemmapView<'a>,
+  anr_trace: Option<MemmapView<'a>>,
+  app_exit_description: Option<&str>,
+  is_file_size_optimization_enabled: bool,
+) -> WIPOffset<v_1::Report<'fbb>> {
+  if let Some(anr_trace) = anr_trace {
```
FranAguilera: To clarify, now platform layer can pass a Optional (nullable) inputTraceStream. If is not null try to parse, if no errors return the full report. If parsing fails (could be a malformed/incomplete input stream raw report) still send a basic report without stack traces

bd-report-parsers/src/android.rs
```diff
@@ -54,11 +54,35 @@ type FrameOffsetSequence = Vec<u32>;
...
+      device_info,
+      anr_trace,
+      app_exit_description,
+      is_file_size_optimization_enabled,
+    )
+  {
+    return report;
+  }
+
+  build_anr_without_trace(builder, app_info, device_info, app_exit_description)
```
FranAguilera, line 79: For the cases where we fail parsing due to incomplete data on the raw report, I still think we should have a general viz count of this error (we have a related project for this, I will add a ticket there and handled then)

snowp: It would be nice if we included the failure in the report somehow but yeah can handle separately

bd-report-parsers/src/android_tests.rs
```diff
@@ -319,25 +319,16 @@ macro_rules! assert_parsed_anr_eq {
...
-      Err(nom::Err::Error(e) | nom::Err::Failure(e)) => {
-        let mut contents = format!("{e:?}");
-        contents.truncate(300);
-        panic!("failed to parse: {contents:?}")
-      },
-      Err(e) => panic!("failed to parse: {e:#?}"),
-    }
+    );
+    let report = get_table!(Report, builder, offset);
+    insta::assert_debug_snapshot!(report);
```
snowp, line 331: Are we missing a checked in snapshot file?

bd-report-parsers/src/android.rs
```diff
@@ -97,7 +134,30 @@ pub fn build_anr<'a, 'fbb>(
...
+  v_1::Report::create(builder, &args)
+}
+
+fn build_anr_without_trace<'fbb>(
+  builder: &mut FlatBufferBuilder<'fbb>,
+  app_info: &v_1::AppMetricsArgs<'fbb>,
+  device_info: &mut v_1::DeviceMetricsArgs<'fbb>,
+  app_exit_description: Option<&str>,
+) -> WIPOffset<v_1::Report<'fbb>> {
+  device_info.platform = v_1::Platform::Android;
```
snowp: nit but seems like an opp place to pin this value, seems like we should do this at the lowest/highest levels of helpers since this should be set for all androi reports?

bd-report-parsers/src/android_tests.rs
```diff
@@ -487,3 +477,52 @@ fn build_anr_without_description_or_subject() {
...
+
+  let report = flatbuffers::root::<Report<'_>>(builder.finished_data()).unwrap();
+  let error = report.errors().unwrap().get(0);
+  assert_eq!(error.name(), Some("User Perceived ANR"));
+  assert_eq!(error.reason(), Some(description));
+  assert_eq!(report.thread_details().unwrap().count(), 0);
+}
+
+#[test]
+fn build_anr_with_unparsable_trace_uses_app_exit_description() -> anyhow::Result<()> {
```
snowp, line 506: might be nice with snapshot tests for more of these just so we can easily see what reports are being genereated by these

snowp (approved):

PR #575, opened 2026-08-07
Add monorepo Bazel integration

snowp:


bd-client-stats/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+rust_library(
+    name = "bd-client-stats",
+    srcs = glob(
+        ["src/**/*.rs"],
+        exclude = ["src/bin/**/*.rs"],
+    ),
+    # logger-cli is a workspace consumer of the optional observation API. Bazel
```
mattklein123, line 12: Isn't this going to cause issues? How do we do features in this world?

logger-cli/src/bin/mcp.rs
```diff
@@ -90,6 +90,7 @@ struct SetFeatureFlagParameters {
 
 struct Tool {
   #[expect(
+    clippy::struct_field_names,
```
mattklein123: I would just allow this globally

fuzz/BUILD.bazel
```diff
@@ -0,0 +1,82 @@
+load("@crates//:defs.bzl", "all_crate_deps")
+load("@rules_rs//rs:rust_library.bzl", "rust_library")
+load("@rules_rs//rs:rust_binary.bzl", "rust_binary")
```
mattklein123: How are you going to handle rewriting the test and binary targets for clippy etc. Should these just point to root paths and not work in isolation or will you rewrite them somehow?

Cargo.toml
```diff
@@ -161,6 +161,7 @@ protobuf-json-mapping = { git = "https://github.com/bitdriftlabs/rust-protobuf.g
 rand         = "0.10.2"
 regex        = "1.13.0"
 reqwest      = { version = "0.13.4", features = ["deflate", "json"] }
+rusqlite     = "0.32.1"
```
mattklein123, line 164: Remove the other dep? This is what valve uses?

mattklein123 (approved):

fuzz/BUILD.bazel
snowp: These should be using the top level ones yeah, let me update. I'd add a lint for that at some point

bd-client-stats/BUILD.bazel
snowp: The only way to do this is to have two distinct build targets, one with and another one without. Downstreams then depend on the one they want

For most crates features are additive and just enables more code so this is fine, but for stuff that changes behavior we'll have to be a bit careful. Transitive features for first party crates isn't really supported in Bazel but it's not that common

I'll clean this up in a follow up to make it so that only logger-cli depends on the extra feature

Cargo.toml
snowp: will do

PR #577, opened 2026-08-07
bd-workflows: implement type-based routing within the workflow engine 

snowp:
Instead of always evaluating all workflows for all logs, this tracks a running set of candidate workflows based
on the log type of the log. This allows us to bypass workflows whose next match transitions require a different
log type than the inbound log.

We keep a router table that provides a routing layer on top of the simple workflow list: for workflows that can be determined to require a particular log type we can filter down the considered workflows to workflows constrained to a log type + all ambiguous workflows. This is kept up to date during workflow execution, rebuilding the routing rules as the list of active runs. 

For now this is only going to apply to initial conditions due to not working with the run timeout defined on most workflows, but the initial run ends up being a very significant part of the workflows considered in practice. 

Fixes BIT-9477

### DIFF
--- a/bd-workflows/src/engine.rs
+++ b/bd-workflows/src/engine.rs
@@ -36,7 +36,7 @@ use crate::config::{
   WorkflowsConfiguration,
 };
 use crate::metrics::MetricsCollector;
-use crate::routing::WorkflowLogRouter;
+use crate::routing::{SelectedWorkflows, WorkflowEventKind, WorkflowLogRouter};
 use crate::sankey_diagram::{self, PendingSankeyPathUpload};
 use crate::workflow::{
   SankeyPath,
@@ -341,7 +341,7 @@ impl<C: CounterTrait, H: HistogramTrait> WorkflowsEngine<C, H> {
     }
 
     self.state.refresh_tracing_counts_from_state();
-    Self::rebuild_log_routes(&mut self.log_router, &self.state.workflows, &self.configs);
+    Self::rebuild_event_routes(&mut self.log_router, &self.state.workflows, &self.configs);
 
     log::debug!(
       "started workflows engine with {} workflow(s); {} pending processing action(s); {} pending \
@@ -470,7 +470,7 @@ impl<C: CounterTrait, H: HistogramTrait> WorkflowsEngine<C, H> {
       .flush_buffers_actions_resolver
       .standardize_streaming_buffers(self.state.streaming_actions.clone());
     // Regenerate routes whenever the workflow configuration is reloaded.
-    Self::rebuild_log_routes(&mut self.log_router, &self.state.workflows, &self.configs);
+    Self::rebuild_event_routes(&mut self.log_router, &self.state.workflows, &self.configs);
 
     log::debug!(
       "consumed received workflows config update; workflows engine contains {} workflow(s)",
@@ -638,8 +638,8 @@ impl<C: CounterTrait, H: HistogramTrait> WorkflowsEngine<C, H> {
       .extend(workflow_tracing_carryover_flush_action_ids);
   }
 
-  /// Rebuilds routes after an event or configuration change that can affect every workflow.
-  fn rebuild_log_routes(
+  /// Rebuilds routes after a configuration change that can affect every workflow.
+  fn rebuild_event_routes(
     log_router: &mut WorkflowLogRouter,
     workflows: &[Workflow],
     configs: &[Config],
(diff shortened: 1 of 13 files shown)


bd-workflows/src/engine.rs
```diff
@@ -796,88 +899,83 @@ impl<C: CounterTrait, H: HistogramTrait> WorkflowsEngine<C, H> {
...
+            index,
+            event,
+            state_reader,
+            now,
+            sampled_roll,
+            &mut workflow_event_output,
+          );
+        }
+      },
+      WorkflowEvent::SessionStart(_) | WorkflowEvent::StateChange(..) => {
```
mattklein123: At least for state change can't we also filter based on whether a workflow cares about it or not? Not sure about SessionStart but I think in that case we could do it also?

bd-workflows/src/routing.rs
```diff
@@ -0,0 +1,210 @@
...
+
+const LOG_TYPE_BUCKET_COUNT: usize = LogType::VALUES.len();
+
+//
+// WorkflowLogRoute
+//
+
+// TODO(snowp): We should be able to apply this routing logic to other event types to dramatically
+// reduce the number of workflows attempted for state events. For now state events are rare so we
+// limit this to logs.
```
mattklein123: I commented on this above.

bd-workflows/src/routing.rs
```diff
@@ -0,0 +1,210 @@
...
+//
+// WorkflowLogRouter
+//
+
+/// Maintains sorted workflow indices for efficient, allocation-free log routing.
+///
+/// This is a derived view over the engine's primary workflow vector: every stored index refers to
+/// the workflow at that same position. The caller must rebuild it after adding, removing, or
+/// reordering workflows; between those boundaries, the router updates only workflows which
+/// processed the current log.
```
mattklein123: nit: not processed, but actual advanced 1 or more traversals?

bd-workflows/src/engine.rs
```diff
@@ -796,88 +899,83 @@ impl<C: CounterTrait, H: HistogramTrait> WorkflowsEngine<C, H> {
...
+    if routed_log {
+      log_router.refresh_selected_routes(|index| {
+        workflows
+          .get(index)
+          .zip(configs.get(index))
+          .map(|(workflow, config)| workflow.log_route(config))
+          .unwrap_or_default()
+      });
+    } else {
+      Self::rebuild_log_routes(log_router, workflows, configs);
```
mattklein123: Per other comments would be nice to now do this always for the other events, even if rare.

bd-workflows/src/engine.rs
```diff
@@ -796,88 +899,83 @@ impl<C: CounterTrait, H: HistogramTrait> WorkflowsEngine<C, H> {
...
-        && !(was_in_initial_state && workflow.is_in_initial_state())
-      {
-        self.needs_state_persistence = true;
-      }
+    let Self {
+      state,
+      configs,
+      stats,
+      needs_state_persistence,
+      log_router,
```
mattklein123, line 926: It seems a little fragile to me that we mutate the router in place and use that mutation for rebuilds, but then could theoretically early exit. Can we somehow debug_assert at least that we finalize anything that should be finalized in a single pass?

bd-workflows/src/engine.rs
```diff
@@ -796,88 +899,83 @@ impl<C: CounterTrait, H: HistogramTrait> WorkflowsEngine<C, H> {
...
-      );
-      tracing_carryover_flush_action_ids.extend(workflow_tracing_carryover_flush_action_ids);
+    if routed_log {
+      log_router.refresh_selected_routes(|index| {
+        workflows
+          .get(index)
+          .zip(configs.get(index))
+          .map(|(workflow, config)| workflow.log_route(config))
+          .unwrap_or_default()
+      });
```
mattklein123: This is not cheap even in the case where nothing advanced. Can we improve the performance of the very common no advance case?

mattklein123 (commented):
How can we understand the perf improvment or not for this in different cases? I think you added a bench can we run it against this change in different scenarios and show the deltas?

snowp:
Running it against the internal corpus I got very good results:

```
origin/main | 114.58 ms
Latest routing changes | 59.82 ms
Change | 47.8% faster
```

This is primarily from this being a low to zero match session. I'll figure out how to include this larger corpus and we could formalize a way to report on workflow engine changes

bd-workflows/src/engine.rs
snowp: Yeah i did this, it was a pretty easy extension

bd-workflows/src/engine.rs
snowp: I tried this out and it seems like for the large corpus that adding an additional value that we track whether we can safely skip refresh ends up being more expensive than just doing it. Tracking the additional value added ~5% over just doing the route rebuild

PR #583, opened 2026-08-10
session: add configuration-based session APIs

snowp:
- Add a configuration-backed session strategy with optional initial ID and inactivity timeout.
- Support explicit session starts with an optional app-provided ID, notifying after durable state is updated.
- Preserve session reporting and previous-process attribution.


mattklein123 (approved):
Yay this is much much better.

PR #590, opened 2026-08-13
add a larger workflow benchmark corpus

snowp:
This now runs by default in the crtierion test runner

The session file is ~38MB uncompressed, so compressing it with zstd to avoid a very large file in git.


mattklein123 (approved):
We could probably figure out a way to put the corpus in S3 if that ends up being easier.

PR #592, opened 2026-08-13
implement platfrom mutex

snowp:
This implements a mutex abstraction that should be preferred when called directly via platform APIs.
It allows us to use os_unfair_lock on iOS which is the preferred mutex type that will provide priority inheritance.

Fixes BIT-9594


mattklein123 (approved):

snowp:
Tested this on iOS with a targeted session API test using tsan, so this seems to be working

PR #597, opened 2026-08-13
bd-session: move towards best effort persistence

snowp:
This updates bd-session to use a single persistence worker task instead which is responsible for persisting the session changes to disk. This means that startNewSession etc will no longer block on session persistence and instead just notifies the persistence thread that the state is dirty. This simplifies the session persistence logic substantially and means that we no longer depend on the ALB to handle session persistence which will be a blocked for the future EventBuffer work. 

We are intentionally weakening the contract between startNewSession and the persistence guarantees here, relying the background task to write to disk and flush()-driven barriers.

Fixes BIT-9297
Fixes BIT-9602

### DIFF
--- a/bd-session/src/lib.rs
+++ b/bd-session/src/lib.rs
@@ -725,6 +725,11 @@ impl Strategy {
     .await;
 
     if let Err(e) = result {
+      // Keep the latest in-memory state dirty so the next flush or mutation retries this
+      // best-effort write. Mutations that occurred during I/O are retained by the same flag.
+      if let Some(state) = self.state.lock().as_mut() {
+        state.persistence_pending = true;
+      }
       on_persistence_failure();
       log::warn!("failed to persist coalesced session snapshot: {e}");
     }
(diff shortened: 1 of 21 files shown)


bd-session/src/lib.rs
```diff
@@ -655,139 +603,141 @@ impl Strategy {
...
     );
 
-    let pending_changed = mutation.persist_pending;
-    let callback = mutation.callback.clone();
-    self.persist_loaded_state(&state, &mutation).await;
-    if pending_changed {
+    if effects.persist {
+      // A full channel already holds a request that will persist the newest in-memory snapshot.
+      let _ignored = self.persistence_tx.try_send(PersistenceRequest::Persist);
+    }
```
mattklein123, line 693: I think it's fine but I would double check that a double persist on an unchanged value works ok. This can race with the worker already seeing the new value I think.

mattklein123 (approved):
This makes sense to me. Let's discuss offline on whether to hold this until we get all the stats changed out.

bd-session/src/lib.rs
snowp: I think this is fine: if there's something in the channel at this point, the worker will eventually re-read the latest state and persist that, so at worst we get double persistence in this case. I think unless we're afraid of someone calling `startNewSession` in a tight loop this is fine

PR #600, opened 2026-08-14
more major client stats changes

mattklein123:
1) Previously, the client stats loop could get hung when an upload is
   submitted to the mux which is not ready. This would prevent all other
   activity including disk flushes, shutdown, etc. This change fully
   decouples preparing uploads and submitting them to API while still
   allowing to handle flushing to disk on request.
2) Previously there was no disk write debounce. This change implements a
   runtime configurable disk debounce which defaults to 1s.
3) Add a bunch of tests and fix some test flakes including simplifying
   some screenshot handling.
4) Remove blocking log functionality entirely. We now only have the flush API.

Fixes BIT-9607
Fixes BIT-9596


AGENTS.md
```diff
@@ -1,9 +1,35 @@
 # Agent Guidelines for Shared Core
 
 - Primary instructions should be sourced from rust.instructions.md
-- ALWAYS use `cargo nextest` commands for verification as it's faster
+- Use the monorepo Bazel wrapper for Rust linting and tests:
```
snowp: imo this should pivot on being in the monorepo or not or it will confuse anyone using this directly, or the monorepo should have this and have language that overrides what the inner agent files specfiy

snowp (approved):

AGENTS.md
mattklein123: OK will just revert this. I agree I'm trying to figure out the best way of making the agent do the right thing.

PR #610, opened 2026-08-24
bd-event-buffer: add a priority-aware log event buffer

snowp:
Adds a bounded, priority-aware logger EventBuffer and shared ingress/control types without changing the existing ALB channel path. It preserves FIFO delivery across retained entries, protects prior-process and state/control work, and resolves terminal completions after unlocking.

The staged migration and remaining ALB wiring are documented in the included plan.


bd-event-buffer/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+//
+// EventBufferEntry
+//
+
+/// A FIFO event-buffer entry.
+///
+/// Ingress events are boxed so control entries do not inherit the larger ingress-event layout in
+/// the backing `VecDeque`.
```
mattklein123: You are going to spend the memory either way, non-boxing might be faster.

bd-event-buffer/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+#[derive(Clone)]
+pub struct EventBuffer {
+  inner: Arc<EventBufferInner>,
+}
+
+struct EventBufferInner {
+  state: PlatformMutex<LoggerEventBufferState>,
+  notify: Notify,
+  #[cfg(test)]
+  waiting_consumers: tokio::sync::watch::Sender<usize>,
```
mattklein123: nit: the pattern I've been using recently is to have an Option<Arc<dyn TestHooks>> which is cleaner and more flexibly from a code perspective. I would see if that can be made to work.

mattklein123 (approved):

PR #632, opened 2026-09-03
Refresh Shared Core dependencies

snowp:
Refresh generated Cargo dependency metadata after the workspace-wide dependency update.


snowp:
Closing to keep the rollout limited to Loop API and Valve.

PR #634, opened 2026-09-09
Retain state when journal persistence fails

snowp:
Retains live state if journaling fails by transitioning the persistent store to bounded in-memory operation at its configured capacity. Workflows and state consumers continue to observe accepted updates during the current process; restart persistence is explicitly degraded and counted.

### DIFF
--- a/bd-versioned-kv/src/versioned_kv_journal/store.rs
+++ b/bd-versioned-kv/src/versioned_kv_journal/store.rs
@@ -261,6 +261,31 @@ impl PendingScopedValues {
   }
 }
 
+//
+// TestHooks
+//
+
+#[cfg(test)]
+trait TestHooks: Send + Sync {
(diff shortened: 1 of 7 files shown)


bd-versioned-kv/src/versioned_kv_journal/store.rs
```diff
@@ -274,6 +276,15 @@ struct PersistentStore {
   max_capacity_bytes: usize,
   // Stats
   stats: CommonStats,
+  // Keeps post-mutation rotation failures deterministic without relying on filesystem faults.
+  #[cfg(test)]
+  fail_next_post_update_rotation: bool,
```
mattklein123: nit: Option<Arc<dyn TestHooks>> will be cleaner here and easier to extend in the future.

mattklein123 (approved):

PR #647, opened 2026-09-21
initial command implementation

mattklein123:
This is not production ready but is enough to start testing. It completely replaces the existing take screenshot workflow action. This PR does not implement the new workflow support. This will be done in a follow up. It also does not have any resilience around command execution across process restarts or mux restarts. All of this will be hardened in a follow up.


bd-artifact-upload/src/uploader.rs
```diff
@@ -468,22 +550,31 @@ impl Uploader {
...
-
-            #[allow(unused)]
-            let name = self.handle_upload_complete().await?;
-
-            #[cfg(test)]
-            if let Some(hooks) = &self.test_hooks {
-                hooks.upload_complete_tx.send(name).await.unwrap();
+            match result? {
+              Ok(()) => {
+                #[allow(unused)]
```
snowp: ? seems used to me

bd-artifact-upload/src/uploader.rs
```diff
@@ -630,6 +712,28 @@ impl Uploader {
     Ok(entry.name)
   }
 
+  async fn discard_upload(&mut self, entry: Artifact, error: String) {
+    if let Err(delete_error) = self
+      .file_system
+      .delete_file(&REPORT_DIRECTORY.join(&entry.name))
```
snowp: should probably rename the REPORT_DIRECTORY constant at least, kinda confusing now that we're putting more stuff in there

bd-buffer/src/buffer/common_ring_buffer.rs
```diff
@@ -718,6 +726,21 @@ impl<ExtraLockedData> LockedData<ExtraLockedData> {
     conditions.next_read_complete.notify_all();
 
     let record_size = reservation.size - guard.extra_bytes_per_record;
+    let on_record_retired_cb = guard.on_record_retired_cb;
+    if on_record_retired_cb(&mut guard.extra_locked_data, record_size).is_err() {
+      // A corrupt in-bounds record length can advance the read pointer before durable accounting
```
snowp, line 731: Is this a new total loss case?

snowp (approved):

bd-buffer/src/buffer/common_ring_buffer.rs
```diff
@@ -718,6 +726,21 @@ impl<ExtraLockedData> LockedData<ExtraLockedData> {
...
+      // rejects it. The remaining buffer state is no longer trustworthy.
+      guard.on_total_data_loss();
```
mattklein123, line 733: Yes. This is a new recovery path introduced in this branch. When durable unread accounting rejects a corrupt in-bounds record length after `advance_next_read`, the state can no longer be trusted. We deliberately drop the buffer contents and return `Unavailable` rather than retaining partially advanced state.

mattklein123 (commented):
Clarified the new total-loss recovery path for corrupted in-bounds record lengths.

